PR #46, opened 2021-05-22
Add Melos

jamesblasco:
Adds melos, format files and updates to last version published


remonh87:
Can you. squash the commits a bit :)?

melos_stripe.iml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
remonh87, line 1: Is this file needed?

stripe/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 1.0.1
+
+ - **FEAT**: first melos version.
```
remonh87, line 3: I do not think this is a user facing feature maybe call it something like: "bug fixes"

remonh87 (approved):
for the rest looks good to me check my comments and then ship it :)

melos_stripe.iml
jonasbark: yeah, should be removed imo

Issue #47, opened 2021-05-22
Fix `flutter analyze` lint issues on the repository


albertus-stripe:
We currently have 815 issues by running `flutter analyze` on the repository.

**To Reproduce**
1. Clone the repository
2. Run `flutter analyze`

**Expected behavior**
We should not be having any lint issues and we should be catching this on a GitHub Action CI and prevent future pull request to be merged if it's violating any lint warnings.

albertus-stripe:
Somewhat related to #42 , we should not be merging any future pull requests that causes any lint warning after this issue is resolved

Issue #42, opened 2021-05-18
Setup CI/CD pipeline


albertus-stripe:
## Problem
We currently don't have a CI/CD pipeline set-up yet for regression testing and automatically building/deploying the package to pub.

I think as the project grows this is going to be important to make sure we are not breaking things and not having any "bus factor" on the project (any maintainer with write access on the repository should be able to easily deploy the package to pub.dev by creating a tag on the Git without too much troubles?)

Just a proposal to make sure that this repository and project scalable in the longer term.

## Solution 
1. CI pipeline
We need to have a CI pipeline for:
- Running automated regression tests so that things don't break after someone makes a pull request
(Though, this is dependant on #19 right now since we don't have good test coverage currently), both for unit tests and integration tests just like flutter/flutter repository.

2. CD pipeline
I think another pipeline that triggers on tag creation to automatically build the binary and package and publish it to pub would also be essential to this repository.

What do you think?
Happy to take this task while also helping with testing and adding automated tests on the project \o/

remonh87:
Good one I think @jonasbark  wanted to setup github actions right? I have also experience with codemagic if this is what we want.

albertus-stripe:
+1  for GitHub Actions since it's more native to GitHub.
I can take this on for GitHub actions if you want!

jamesblasco:
Same here! Let's do Github Actions

@albertus-stripe  Let us know how you plan to start it and how we can help. We can separate into smaller tasks and divide them between the three of us 

For CD I think we could use melos. It is a very useful tool https://docs.page/invertase/melos/automated-releases

jonasbark:
It's indeed on my todo after the next release with the payment sheet - I sure would love some help there :) 

The focus could we the CD part, with the CI one (formatting, automated tests) being second

albertus-stripe:
Reopening the issue since it got closed automatically by the PR on the CI work (my mistake for linking it on the PR, sorry!)
I'd love to help with the CD pipeline as well!
Since the melos work now has been finished. Will continue the discussion offline in our Slack!

albertus-stripe:
Hey maintainers! @jonasbark @remonh87 @jamesblasco 
I have another (hopefully final) nitpick suggestion regarding the CD pipeline and how we do releases for this package.
As I'm not really aware with how you guys do the current release process, I highly recommend sticking to Git tag versioning for all releases going forward.
[...]

PR #48, opened 2021-05-22
Fix static analysis issues, create CI pipeline using GitHub Actions, move example folder to root level

albertus-stripe:
## Related Issues
#42 #47 

## Summary
* Create reusable analysis_options.yaml for all plugins based on pedantic
* Fix all static analysis issues
* Create CI using GitHub actions for continuous build and static analysis
* Move `stripe/example` folder to root level

## Next Steps
- Update GitHub actions scripts once melos PR is merged
- Document all public members and use stricter analysis_options.yaml (optional / need to be discussed with the team)

### analysis_options.yaml
Previous analysis_options.yaml seems to be based on the [flutter/flutter](github.com/flutter/flutter) repository which can be too strict for us (like always inferring the type of any variables, for example), this makes sense on flutter/flutter repository since they are constantly using the latest Dart SDK which can be broken at any moment. In my personal opinion, using `pedantic` would keep the code quality of our codebase remains high without being too strict on the linter issues.

We should be either using the `pedantic` settings which is The Dart analyzer settings and best practices used internally at Google, or, we can also customize the linter ourselves.
Happy to discuss and be flexible on this matter.

### Moving `stripe/example` to root level
There are several reasons why I need to do this:
- We had the CI failing when running `flutter analyze` because the `.env.dart` was missing (since the template is `.env.example.dart`)
- To increase the visibility of the example on the project. Many open source project also adopts the same folder structure.


.github/workflows/all_plugins.yaml
```diff
@@ -0,0 +1,60 @@
+name: all_plugins
```
albertus-stripe, line 1: I think now would be a good time to add the GitHub actions CI now after fixing all the analyzer error/violations to make sure that future PR does not add any further violations/error which decreases the quality of our codebase!

stripe/.idea/libraries/Dart_SDK.xml
```diff
@@ -1,19 +0,0 @@
...
-      <root url="file:///Users/jaimeblasco/Documents/Projects/flutter/bin/cache/dart-sdk/lib/io" />
-      <root url="file:///Users/jaimeblasco/Documents/Projects/flutter/bin/cache/dart-sdk/lib/isolate" />
-      <root url="file:///Users/jaimeblasco/Documents/Projects/flutter/bin/cache/dart-sdk/lib/math" />
-      <root url="file:///Users/jaimeblasco/Documents/Projects/flutter/bin/cache/dart-sdk/lib/mirrors" />
-      <root url="file:///Users/jaimeblasco/Documents/Projects/flutter/bin/cache/dart-sdk/lib/typed_data" />
-    </CLASSES>
-    <JAVADOC />
-    <SOURCES />
-  </library>
-</component>
```
albertus-stripe, line 19: Just deleting all .idea folder from the Git since seems like previously this was not gitignored properly!

analysis_options.yaml
```diff
@@ -0,0 +1,21 @@
+# This is the default analysis_options.yaml for all packages and examples
```
albertus-stripe, line 1: Putting this at root level to have a default and shared analysis_options for all the 4 packages in this repository!

analysis_options.yaml
```diff
@@ -0,0 +1,21 @@
...
+# If a package needs specific rules, create the analysis_options.yaml on it
+# include this file, like:
+#
+# include: ../analysis_options.yaml
+#
+# TODO (albertus-stripe): Change this to all_lint_rules.yaml for stricter lint and better codebase going forward after all documentation on public members are done.
+include: package:pedantic/analysis_options.1.11.0.yaml
```
albertus-stripe, line 8: I think pedantic options is quite a good one here, ensuring the code quality of our codebases remains high without being too strict and slowing us down in the long term. But happy to discuss about this changes if any of you have different opinion

analysis_options.yaml
remonh87: yes good thinking 👍

example/lib/screens/apple_pay_screen.dart
```diff
@@ -3,7 +3,6 @@ import 'dart:convert';
 import 'package:flutter/material.dart';
 import 'package:http/http.dart' as http;
 import 'package:flutter_stripe/flutter_stripe.dart';
-import 'package:stripe_example/widgets/loading_button.dart';
```
remonh87, line 6: In general I am not convinced of moving the example to it's own folder. If I look to the federated plugins of google they all put the example inside the main dart package. See for example https://github.com/flutter/plugins/tree/master/packages/url_launcher . I think it will be confusing where the example is for in this case.
- [ ] Revert moving the example out of the stripe package

stripe/pubspec.lock
```diff
@@ -145,9 +152,9 @@ packages:
   stripe_ios:
     dependency: "direct main"
     description:
-      name: stripe_ios
-      url: "https://pub.dartlang.org"
-    source: hosted
+      path: "../stripe_ios"
```
remonh87, line 155: I think this will cause problems when we are publishing 

- [ ]  Check when running `pub publish --dry run`

.github/workflows/all_plugins.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+on:
```
remonh87, line 3: Add a step for testing publishing.  In my other package we found this check quick and very helpful in case we messed up with the project structure.  Also I think we should consider building the apps (can be in a pr later of course)  since we have quite some native code.

- [ ] Create job which runs `pub publish --dry-run` for packages

remonh87 (changes requested):
Looks great and would be a big quality improvement for the package. just would like to see the suggested points and then I will approve!

example/lib/screens/apple_pay_screen.dart
jonasbark: I would think it is easier to find like that

albertus-stripe: Hey! @remonh87
Yes, for the Flutter official plugins, I think it makes a lot of sense to put the example folder inside the Dart package. But that is because the [`flutter/plugins`](https://github.com/flutter/plugins) is a mono-repo for multiple published plugins with different functionalities.

But if you notice the [`flutter/flutter`](https://github.com/flutter/flutter), they actually have the `examples` folder in the root level, containing multiple examples for the Flutter repository which in my opinion helps a lot since it's for folks to discover the `example` folder as well (I doubt a lot of people will go to the `stripe` folder unless they are interested in contributing or we mention it explicitly on the README.md).

I think putting the `example` in the `stripe` folder would be better if we plan to have a separate example for each of the folder: `stripe_android`, `stripe_ios`, and `stripe_platform_interface`. Are we planning to do that?
Because currently it's not obvious where the example is located in the codebase if I'm particularly just a user interested in using this package and reading the README.md of this repository.

Since technically this repository should contain one functionality: the Stripe SDK, I think it's much better to have the `example` in the root level, following the example of the [Stripe React Native SDK repository](https://github.com/stripe/stripe-react-native).

Another reason why I moved the `example` is because of https://github.com/flutter-stripe/flutter_stripe/runs/2647177553 in this PR was failing because of an error in the `example` folder due to the `.env.dart` missing in the plugin.. Running `flutter analyze` on the `stripe` folder would run the analyzer twice: one on the `example` folder, and then another in the current `stripe` package codebase.
Any suggestion on how to resolve this issue if you have a strong opinion to still keep this `example` folder in `stripe` level?

@jonasbark easier to find like that meaning in the `stripe` folder or in the root level?

jonasbark: I'd definitely vote for it to be in the root folder. 
But no separate example per platform plugin

albertus-stripe: Ah, I see. Thanks, @jonasbark !
Yeah, I'm also in strong favor of having the example in the root folder for better visibility as well, since I don't think we can also have a useful example per platform plugin?

remonh87: I understand the rationale my comment is more that we make a conscious decision to deviate. We should make sure that the example is correctly referred to in pub.dev

remonh87 (approved):

example/lib/screens/apple_pay_screen.dart
albertus-stripe: definitely, we should make sure to set the example correctly on the `pub.dev` website as well.
Sorry @remonh87, I'm still quite unclear here, so are you still in favor of reverting the changes on keeping the `example` on `stripe` or would you be ok with moving this `example` on the root level here?

Would also love to hear your opinion since I know you have been working and changing a lot of the `example` code, want to make sure this change is ok with you! @jamesblasco

stripe/pubspec.lock
albertus-stripe: Yes, you are right, this will cause problems when we are publishing.
But currently it's not only the `stripe_ios` causing the problem, previous `pubspec.lock` that was checked in had dependencies on the `stripe_android` and `stripe_platform_interface` as well.
I notice that other Flutter plugins don't check-in their `pubspec.lock` on the Git, so I think I'll create an issue and do it in a separate PR if you don't mind @remonh87

.github/workflows/all_plugins.yaml
albertus-stripe: Thanks Remon, will do this in a separate PR and created this issue like we discussed offline
https://github.com/flutter-stripe/flutter_stripe/issues/52

stripe/pubspec.lock
albertus-stripe: https://github.com/flutter-stripe/flutter_stripe/issues/51

Added this issue and will do it in a separate PR as we discussed offline ^^

analysis_options.yaml
jamesblasco: I would personally go for the new https://github.com/dart-lang/lints package.

example/lib/screens/apple_pay_screen.dart
jamesblasco: I would also prefer to have it at the root level. 

And for pub.dev is better, right now the repo link is `https://github.com/flutter-stripe/flutter_stripe/tree/main/stripe` to allow the example to appear. With this change, we can have `https://github.com/flutter-stripe/flutter_stripe/`

.github/workflows/all_plugins.yaml
jamesblasco: Let's add it once melos is merged

jamesblasco:
Mmmh I feel this PR is getting too big. I would prefer if we can separate
 - Fix analysis issues
 - Move packages to packages/ and example to root level
 - create CI pipeline
If the commit history is a bit messy let's squash and we move the packages in another PR

jonasbark:
no need to separate it now IMO, don't see an advantage at this stage

example/lib/screens/apple_pay_screen.dart
albertus-stripe: awesome, will resolve this conversation.

analysis_options.yaml
albertus-stripe: @jamesblasco Yes sure! I think pedantic would be a subset of this package. Would you be ok with us temporarily using `pedantic` for now and re-evaluating this once we have agreement and decision on which lint?
Changing this on this PR would make the PR even bigger than it is right now.

albertus-stripe:
Thanks! @jonasbark 
Yeah, separating now also requires me to redo some of the works on this PR :(
I'll keep this in mind for future PRs and make it much smaller.

Would you be ok with letting this pass for now? @jamesblasco @jonasbark

jonasbark (approved):

PR #50, opened 2021-05-23
Improve code sync

jonasbark:
This will ensure code can be synced without any adjustments needed - the only exception is two castings in StripeSdkCardViewManager. Don't think there's a way to change this.


jamesblasco:
Mmm could you rebase main into this branch? The commit history is a bit messed

jonasbark:
I'll squash it instead

PR #62, opened 2021-05-24
add integration test infrastructure

jonasbark:
TODO:

- [x] blocked by not being able to set credit card details
- [x] test on Android and iOS, Firebase Test Lab
- [x] decide on whether to do widgets tests (DRY) or test the plugin interface only


albertus-stripe:
exciting work, just papertrailing here to follow on the progress 👀

jamesblasco:
Great work @jonasbark ! I feel FirebaseTestLab can become a bit expensive for daily use (free 10 uses per day).
We could try using simulators in Github actions for the PR checks and FirebaseTestLabs daily checks with the new github cron feature
There is an example using simulators: https://github.com/FirebaseExtended/flutterfire/blob/master/.github/workflows/scripts/drive-example.sh

.github/workflows/all_plugins.yaml
```diff
@@ -73,8 +71,16 @@ jobs:
...
 
+
+    - name: "Start dev server"
+      run: cd example/server/ && npm install && (npm run start:dev &) && cd ..
+
+    - name: "Set IP address"
+      run: echo "const kApiUrl='$(ifconfig | grep -Eo 'inet (addr:)?([0-9]*\.){3}[0-9]*' | grep -Eo '([0-9]*\.){3}[0-9]*' | grep -v '127.0.0.1')';" > integration_test/ip.dart
+
     - name: "Run test"
-      run: melos run test
```
jamesblasco, line 80: Leave 
```yaml
- name: "Run test"      
   run: melos run test;
```
for the unit tests before the integration tests (commented if you want for now)

jonasbark: tried, doesn't run the correct flutter drive command out of the box

jamesblasco: Yeah, I mean separate it into two 
- name: "Run unit test"      
   run: melos run test;
- name: "Run integration test"      
   run: flutter run .....;

Check the melos.yaml file in the root level. I defined the test command myself. But for integration test we don't need melos as we will only test `flutter_stripe` no?

Issue #72, opened 2021-05-25
Undefined behavior (app crash) with Google Pay on Android


anmentone:
When using the GooglePayButton widget I get strange behavior if I decide to cancel the payment by closing the open modal after pressing the pay button. 

**To reproduce on example app**
I wanted to make a video demonstration but unfortunately, the google play screen is obscured (obviously) with any screen capture. 
The steps are pretty simple:
1. On the example App, choose **Google Pay payment (Android)**
2. Then press the pay button
3. Close the Google Pay screen, without making any payment. I've tried the X in the top right corner and the back button on the navigation bar of my Huawei phone and the app crashes trying to close the screen.

The error messages seem to be clear, I'm trying to take a deeper look.
<details>
<summary>log: RuntimeException</summary>

```
D/AndroidRuntime(19083): Shutting down VM
I/QarthLog(19083): [PatchStore] createDisableExceptionQarthFile
I/QarthLog(19083): [PatchStore] create disable file for com.flutter_stripe.example uid is 10473
E/AndroidRuntime(19083): FATAL EXCEPTION: main
E/AndroidRuntime(19083): Process: com.flutter_stripe.example, PID: 19083
E/AndroidRuntime(19083): java.lang.RuntimeException: Failure delivering result ResultInfo{who=null, request=991, result=0, data=null} to activity {com.flutter_stripe.example/com.flutter_stripe.example.MainActivity}: java.lang.NullPointerException: Parameter specified as non-null is null: method kotlin.jvm.internal.Intrinsics.checkNotNullParameter, parameter data
E/AndroidRuntime(19083): 	at android.app.ActivityThread.deliverResults(ActivityThread.java:5471)
E/AndroidRuntime(19083): 	at android.app.ActivityThread.handleSendResult(ActivityThread.java:5512)
E/AndroidRuntime(19083): 	at android.app.servertransaction.ActivityResultItem.execute(ActivityResultItem.java:51)
E/AndroidRuntime(19083): 	at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:149)
E/AndroidRuntime(19083): 	at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:103)
E/AndroidRuntime(19083): 	at android.app.ActivityThread$H.handleMessage(ActivityThread.java:2386)
E/AndroidRuntime(19083): 	at android.os.Handler.dispatchMessage(Handler.java:107)
[...]

PR #74, opened 2021-05-25
Fix intent error according to Stripe RN plugin

jamesblasco:
Fixes https://github.com/flutter-stripe/flutter_stripe/issues/72

According to the latest version Intent is nullable

https://github.com/stripe/stripe-react-native/blob/7b02f143f5d72fcd14d9acd8ab93570f2853afed/android/src/main/java/com/reactnativestripesdk/StripeSdkModule.kt#L46

### DIFF
--- a/packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/StripeSdkModule.kt
+++ b/packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/StripeSdkModule.kt
@@ -41,7 +41,7 @@ class StripeSdkModule(reactContext: ReactApplicationContext, cardFieldManager: S
   private var confirmPaymentClientSecret: String? = null
 
   private val mActivityEventListener = object : BaseActivityEventListener() {
-    override fun onActivityResult(activity: Activity, requestCode: Int, resultCode: Int, data: Intent) {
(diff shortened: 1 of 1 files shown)


packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -41,7 +41,7 @@ class StripeSdkModule(reactContext: ReactApplicationContext, cardFieldManager: S
   private var confirmPaymentClientSecret: String? = null
 
   private val mActivityEventListener = object : BaseActivityEventListener() {
-    override fun onActivityResult(activity: Activity, requestCode: Int, resultCode: Int, data: Intent) {
+    override fun onActivityResult(activity: Activity, requestCode: Int, resultCode: Int, data: Intent?) {
```
remonh87, line 44: are you sure that intent should be nullable?  I do not see this at the original implementation of react-native. https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/bridge/BaseActivityEventListener.java

@jonasbark  can you see some potential issues here?

remonh87 (approved):

PR #84, opened 2021-05-27
allow url scheme customized from flutter

dariuspo:


packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/StripeSdkModule.kt
```diff
@@ -219,7 +219,7 @@ class StripeSdkModule(reactContext: ReactApplicationContext, cardFieldManager: S
     val stripeAccountId = getValOr(params, "stripeAccountId", null)
     val urlScheme = getValOr(params, "urlScheme", null)
     val setUrlSchemeOnAndroid = getBooleanOrFalse(params, "setUrlSchemeOnAndroid")
```
jonasbark, line 221: I think what you really want is to add setUrlSchemeOnAndroid  in stripe.dart and get rid of the change below

packages/stripe/pubspec.yaml
```diff
@@ -11,9 +11,13 @@ environment:
 dependencies:
   flutter:
     sdk: flutter
-  stripe_android: ^1.0.1+3
+  stripe_android:
```
jonasbark, line 14: this should be reverted

remonh87:
@dariuspo  are you still planning to update the pr?  I think if you address @jonasbark  comments we can integrate it.

remonh87:
thanks for your contribution we incorporated the changes in the `sync` branch. It will be part of the new release,  pr can be closed now.

PR #107, opened 2021-06-01
Rename PaymentMethodParams.aliPay to PaymentMethodParams.alipay for better consistency

albertus-stripe:
Alipay is a digital wallet in China that has more than a billion active users worldwide.

The word Alipay is a single world, so we should not be using `camelCase` in this case.

This is a breaking change but it's much better to do this early in the beta than having the user of this package being confused after the launch.

Nitpick and very small changes :)


jonasbark (approved):

remonh87:
Wait with merging I will add already the breaking change to the changelog so we do not forget

- [ ] add breaking change to changelog

albertus-stripe:
hey @remonh87,
just a gentle bump on the PR, can we merge this now?

PR #118, opened 2021-06-04
export config params for apple pay button

remonh87:


example/lib/screens/apple_pay_screen.dart
```diff
@@ -40,6 +40,7 @@ class _ApplePayScreenState extends State<ApplePayScreen> {
               padding: EdgeInsets.all(16),
               child: ApplePayButton(
                 onPressed: _handlePayPress,
+                type: ApplePayButtonType.buy,
```
remonh87, line 43: for now also added this to prevent us from happening again

jamesblasco (approved):

PR #144, opened 2021-06-13
Fix https://github.com/flutter-stripe/flutter_stripe/issues/14

jonasbark:
Uses reflection - applies the workaround if an older Flutter version is used


packages/stripe/README.md
```diff
@@ -35,7 +46,12 @@ dart pub add flutter_stripe
 
 #### Android
 
-Android 5.0 (API level 21) and above
+- Android 5.0 (API level 21) and above
```
remonh87, line 49: Let's also add the gradle wrapper version here.

remonh87 (approved):

PR #158, opened 2021-06-20
add missing payment methods, fixes #106

### LINKED ISSUE #106: Missing PaymentMethodParams for other payment methods
Looking at `stripe_platform_interface/lib/src/models/payment_methods.dart`, seems like currently we are not having the `PaymentMethodParams` for these payment methods:
- OxxoParams
- Sofort
- SepaDebit
- AfterpayClearpay
- Eps
- AuBecsDebitParams
- FPX (created a PR for this)

We should create `PaymentMethodParams` for these payment methods following the React Native types declaration here:
https://github.com/stripe/stripe-react-native/blob/master/src/types/PaymentMethods.ts

jonasbark:


packages/stripe_platform_interface/lib/src/models/payment_methods.dart
```diff
@@ -383,6 +383,42 @@ class PaymentMethodParams with _$PaymentMethodParams {
...
   }) = _PaymentMethodParamsFpx;
 
+  @JsonSerializable(explicitToJson: true)
+  @FreezedUnionValue('SepaDebit')
+  const factory PaymentMethodParams.sepaDebit({
+    required String iban,
+    PaymentIntentsFutureUsage? setupFutureUsage,
+
+    /// Billing information.
+    required BillingDetails billingDetails,
```
remonh87, line 393: are you sure that billingdetails are required? I think for sepa they can be nullable.

remonh87 (approved):
nice! happy we got some new methods in looks good to me can you just double check for the payment methods if `Billingdetails` are nullable

PR #162, opened 2021-06-23
Sync with latest sources

jonasbark:


remonh87 (approved):
looks good to me. Did some additional tests and it works

PR #199, opened 2021-07-11
implement result for paymentsheet

### LINKED ISSUE #184: how to handle payment failed on close icon click of card payment ?
**Describe the bug**

**To Reproduce**
1. Entering number 123-456-789 into the cardfield.
2. Tapping the confirm button.
3. Observe a failure with exception (including the part of the stack trace, belonging to this package) …

**Expected behavior**

**Smartphone / tablet**
 - Device: [e.g. iPhone X]
 - OS: [e.g. iOS 13, Android 10]
 - Package version: [e.g. 1.0.0]
 - Flutter version [e.g. 2.0.0.]

**Additional context**

remonh87:
Fix #184

### DIFF
--- a/packages/stripe_ios/ios/Classes/StripeSdk.swift
+++ b/packages/stripe_ios/ios/Classes/StripeSdk.swift
@@ -169,7 +169,7 @@ public class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSele
                 self.paymentSheet?.present(from: UIApplication.shared.delegate?.window??.rootViewController ?? UIViewController()) { paymentResult in
                     switch paymentResult {
                     case .completed:
-                        resolve([String: String]())
+                        resolve([])
                     case .canceled:
                         resolve(Errors.createError(PaymentSheetErrorType.Canceled.rawValue, "The payment has been canceled"))
                     case .failed(let error):
--- a/packages/stripe_ios/ios/Classes/StripeSdk.swift
+++ b/packages/stripe_ios/ios/Classes/StripeSdk.swift
@@ -169,7 +169,7 @@ public class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSele
                 self.paymentSheet?.present(from: UIApplication.shared.delegate?.window??.rootViewController ?? UIViewController()) { paymentResult in
                     switch paymentResult {
                     case .completed:
-                        resolve([])
+                        resolve([String: String]())
                     case .canceled:
                         resolve(Errors.createError(PaymentSheetErrorType.Canceled.rawValue, "The payment has been canceled"))
                     case .failed(let error):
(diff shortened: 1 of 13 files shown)


remonh87:
I need to add a test to it

packages/stripe_platform_interface/lib/src/models/payment_sheet.dart
```diff
@@ -102,3 +102,30 @@ class PresentPaymentSheetParameters with _$PresentPaymentSheetParameters {
   factory PresentPaymentSheetParameters.fromJson(Map<String, dynamic> json) =>
       _$PresentPaymentSheetParametersFromJson(json);
 }
+
+@freezed
+class PaymentSheetResult with _$PaymentSheetResult {
```
jonasbark, line 107: it seems to be a general error, not only targeted to payment sheet so I would advise to rename it 
https://github.com/stripe/stripe-react-native/blob/master/ios/StripeSdk.swift#L183

packages/stripe_ios/ios/Classes/StripeSdk.swift
```diff
@@ -169,7 +169,7 @@ public class StripeSdk: RCTEventEmitter, STPApplePayContextDelegate, STPBankSele
                 self.paymentSheet?.present(from: UIApplication.shared.delegate?.window??.rootViewController ?? UIViewController()) { paymentResult in
                     switch paymentResult {
                     case .completed:
-                        resolve([])
+                        resolve([String: String]())
```
jonasbark, line 172: can you revert this to not divert from react native sources?

remonh87: the problem is that React native is not consistent. On Android they return an empty map and on iOS an empty list. I can do the conversion in dart layer without issues but it will be a bit ugly

packages/stripe_platform_interface/lib/src/models/payment_sheet.dart
remonh87: yes I saw this as well am already fixing it into a class called `LocalizedErrorMessage`

remonh87:
@jonasbark  thanks for the review I adressed all the remarks can you check again?

jonasbark (approved):

hockleongloh:
Hi, want to ask when there will be a new version release for this fixes? Want to integrate this in my project

jacopofranza:
Hello, I want to ask too when will be a new version for this fixes? In 2.0.2 these are no fixed yet.

remonh87:
the changes are that when things are going well we return void and when we the sheet gets cancelled or has an error we will throw a `StripeException` this behaviour is consistent with the stripe native sdk

jacopofranza:
@remonh87 thanks for the quick reply, but this seems not to be the current behaviour, if I use customFlow = true, how can I know if user has intentionally selected a card to proceed or closed the bottomsheet, before call the confirmation?
When I call Stripe.instance.confirmPaymentSheetPayment() and close the bottom sheet not exception are thrown.

Issue #207, opened 2021-07-14
type 'Null' is not a subtype of type 'Map<String, dynamic>' in type cast 


MadhanAD:
**Describe the bug**
From [this example](https://github.com/flutter-stripe/flutter_stripe/blob/main/example/lib/screens/webhook_payment_screen.dart) I was trying to integrate Stripe payment.

Getting null exception while calling `Stripe.instance.confirmPaymentMethod('clientSecret'
      PaymentMethodParams.card(
        billingDetails: billingDetails
      ))`

**Error log while calling confirmPaymentMethod()**
```
[VERBOSE-2:ui_dart_state.cc(199)] Unhandled Exception: type 'Null' is not a subtype of type 'Map<String, dynamic>' in type cast
#0      MethodChannelStripe.confirmPaymentMethod (package:stripe_platform_interface/src/method_channel_stripe.dart:89:44)
<asynchronous suspension>
#1      Stripe.confirmPaymentMethod (package:flutter_stripe/src/stripe.dart:219:29)
<asynchronous suspension>
#2      _HomeScreenState._handlePayPress (package:flutter_base_app/screens/android/HomeScreen.dart:99:27)
<asynchronous suspension>
```

**Smartphone**
 - Device: [e.g. iPhone 12 simulator]
 - OS: [e.g. iOS 14.5]
 - Package version: [e.g. 1.1.0]
 - Flutter version [e.g. 2.2.1]

jonasbark:
This looks very similar to https://github.com/flutter-stripe/flutter_stripe/issues/206
can you try the solution from that ticket?

Gonna assign this to @remonh87 - we should handle it in Dart

MadhanAD:
@jonasbark I have added a publishable key in my main.dart as shown in the example.
Still having the same issue.

jonasbark:
Please provide an example project / code so we can reproduce it

MadhanAD:
@jonasbark  Here is the [Sample Project](https://github.com/MadhanAD/flutter_base) to reproduce the issue

MadhanAD:
@jonasbark  do you have any updates for me?
Did you find any issue on this code?

remonh87:
I am working on it at the moment expect to have it fixed this week.

Issue #208, opened 2021-07-14
confirmPaymentMethod not returning error


tudor07:
1. Use `confirmPaymentMethod` with a card that will generate some error (example: `4000000000009995` for insufficient funds)

Expected result:
Error is returned with the right error code.

Actual result:
Exception is thrown:
```
Unhandled Exception: Assertion failed: "Result should not be null"
#0      UnfoldToNonNull.unfoldToNonNull
package:stripe_platform_interface/src/method_channel_stripe.dart:235
#1      MethodChannelStripe.confirmPaymentMethod
package:stripe_platform_interface/src/method_channel_stripe.dart:91
<asynchronous suspension>
#2      Stripe.confirmPaymentMethod
package:flutter_stripe/src/stripe.dart:219
```

The problem is here: `stripe_platform_interface-1.1.0/lib/src/method_channel_stripe.dart` at line 91.
The `result` variable contains no `paymentIntent`, but this library tries to parse it even if it's null, hence the exception being thrown.
```
final tmp = result?['paymentIntent'] as Map<String, dynamic>;

return PaymentIntent.fromJson(tmp.unfoldToNonNull());
```

The `result` variable contains `error` field with `declineCode`, `stripeErrorCode ` and other useful stuff. There should be a `PlatformException` thrown instead that contains the error codes and everything.

PR #209, opened 2021-07-15
Implement proper error messages

remonh87:
Idea is for every operation:

- Success => we return the result object
- Failure on the Stripe SDK that the user should handle (e.g. a card get rejected, payment sheet gets cancelled). => throw StripePlatformException with the StripeLocalizedErrorMessage that we receive from the SDK.
- In case something unforseen happens in the parsing of the result (e.g. dart is not in sync) => throw StripeError , the user should not catch this.


packages/stripe_platform_interface/lib/src/models/errors.dart
```diff
@@ -30,6 +30,21 @@ T _dataFromJson<T>(Map<String, dynamic> input) => input['code'] as T;
 Map<String, dynamic> _dataToJson<T>(T input) => {'code': input};
 
 @freezed
+
+/// Exception retrieved from the Stripe platform.
+class StripePlatformException with _$StripePlatformException {
```
jonasbark, line 35: I'd prefer StripeError - "Stripe platform" sounds somewhat misleading

packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
```diff
@@ -228,18 +233,20 @@ class MethodChannelStripe extends StripePlatform {
...
     if (result != null) {
       if (result.isEmpty) {
-        return const PaymentSheetResult.success();
+        return;
       } else {
         if (result['error'] != null) {
           //workaround for tojson in sumtypes
           result['runtimeType'] = 'failed';
-          return PaymentSheetResult.fromJson(result);
+          return ResultParser<void>(parseJson: (_) => {})
```
jonasbark, line 244: this seems weird - why parse at all then?

packages/stripe_platform_interface/lib/src/models/errors.dart
remonh87: I wouldn't call it then `StripeError` because of the following:
- error is a problem that the user should not catch , most of the time something unforseen happens.
-  exception is a problem that the user might want to catch.

agree?

packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
remonh87: I agree will fix it

packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
```diff
@@ -78,22 +79,22 @@ class MethodChannelStripe extends StripePlatform {
...
-
-      return PaymentIntent.fromJson(tmp.unfoldToNonNull());
-    } on Exception catch (_) {
+    final result = await _methodChannel
+        .invokeMapMethod<String, dynamic>('confirmPaymentMethod', {
+      'paymentIntentClientSecret': paymentIntentClientSecret,
+      'params': params.toJson(),
+      'options': options,
+    });
+    if (result == null) {
```
jonasbark, line 88: can this ever happen? I don't think we should handle this specifically

remonh87: it is very theoretical. It doesn't hurt :)

remonh87:
Fix #207 #208

packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
```diff
@@ -78,22 +85,22 @@ class MethodChannelStripe extends StripePlatform {
...
       throw const StripeError<PaymentIntentError>(
+        message:
+            "Result was not expected to be null this is probably a error in the plugin",
```
jonasbark, line 97: I'd much rather remove it - otherwise correct 'a error' to 'an error'

packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
```diff
@@ -127,20 +144,21 @@ class MethodChannelStripe extends StripePlatform {
...
         code: PaymentIntentError.unknown,
-        message: 'Handle  payment intent for card failed',
       );
+    } else {
+      return ResultParser<PaymentIntent>(
+              parseJson: (json) => PaymentIntent.fromJson(json))
+          .parse(result: result, succesResultKey: 'paymentIntent');
```
jonasbark, line 161: see the warning below

remonh87:
Fixed last comments

remonh87:
@jamesblasco  would like to hear your opinion about this as well

MadhanAD:
@remonh87 If this issue was fixed, then could you merge this.

remonh87:
First @jamesblasco  and @jonasbark  need to submit their verdict

packages/stripe_platform_interface/lib/src/result_parser.dart
```diff
@@ -0,0 +1,34 @@
+import 'package:stripe_platform_interface/stripe_platform_interface.dart';
+
+class ResultParser<T> {
+  const ResultParser({
+    required T Function(Map<String, dynamic>) parseJson,
+  }) : _parseJson = parseJson;
+
+  T parse(
+      {required Map<String, dynamic> result, required String successResultKey}) {
+    final succesResponse = result[successResultKey];
```
jamesblasco, line 10: LGTM, only fix this typo needs to be fixed.

jamesblasco (approved):
LGTM.

packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
```diff
@@ -127,21 +122,15 @@ class MethodChannelStripe extends StripePlatform {
...
-    }
+    final result = await _methodChannel
+        .invokeMapMethod<String, dynamic>('handleCardAction', {
+      'paymentIntentClientSecret': paymentIntentClientSecret,
+    });
+
+    final tmp = ResultParser<PaymentIntent>(
+            parseJson: (json) => PaymentIntent.fromJson(json))
+        .parse(result: result!, successResultKey: 'paymentIntent');
+    return tmp;
```
jonasbark, line 133: let's get rid of all 'tmp' variables throughout the file, also result shouldn't be forced to be not null

jonasbark (approved):

PR #212, opened 2021-07-16
remove handlecardcation promise to prevent crash

dariuspo:
demo
crash after pay with credit card then grab pay vice versa

crash log:
E/AndroidRuntime(26127): FATAL EXCEPTION: main
E/AndroidRuntime(26127): Process: com.gankglobal.gank, PID: 26127
E/AndroidRuntime(26127): java.lang.IllegalStateException: Reply already submitted
E/AndroidRuntime(26127): 	at io.flutter.embedding.engine.dart.DartMessenger$Reply.reply(DartMessenger.java:155)
E/AndroidRuntime(26127): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler$1.success(MethodChannel.java:238)
E/AndroidRuntime(26127): 	at com.facebook.react.bridge.Promise$1.run(Promise.java:25)
E/AndroidRuntime(26127): 	at android.os.Handler.handleCallback(Handler.java:938)
E/AndroidRuntime(26127): 	at android.os.Handler.dispatchMessage(Handler.java:99)
E/AndroidRuntime(26127): 	at android.os.Looper.loop(Looper.java:223)
E/AndroidRuntime(26127): 	at android.app.ActivityThread.main(ActivityThread.java:7656)
E/AndroidRuntime(26127): 	at java.lang.reflect.Method.invoke(Native Method)
E/AndroidRuntime(26127): 	at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:592)
E/AndroidRuntime(26127): 	at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:947)
I/Process (26127): Sending signal. PID: 26127 SIG: 9

crash happen after we use credit card that have 3d secure, 
then we use other confirm payment method channel(grabpay) or vice versa.

### DIFF
--- a/packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/StripeSdkModule.kt
+++ b/packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/StripeSdkModule.kt
@@ -31,7 +31,6 @@ class StripeSdkModule(reactContext: ReactApplicationContext, cardFieldManager: S
 
   private var urlScheme: String? = null
   private var confirmPromise: Promise? = null
-  private var handleCardActionPromise: Promise? = null
   private var confirmSetupIntentPromise: Promise? = null
   private var confirmPaymentSheetPaymentPromise: Promise? = null
   private var presentPaymentSheetPromise: Promise? = null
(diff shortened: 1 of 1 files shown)


remonh87:
I think we should not do this since we get this part from the ReactNative sdk. Instead if this cause a crash let's log an issue for Stripe in order to fix it. @jonasbark  what do you think?

jonasbark:
Yes, let's wait for the sync in the sync branch first

PR #233, opened 2021-07-25
Add support for fontFamily on iOS

jamesblasco:


packages/stripe_ios/ios/Classes/CardFieldView.swift
```diff
@@ -216,13 +216,39 @@ class CardFieldView: NSObject, FlutterPlatformView, STPPaymentCardTextFieldDeleg
...
+                cardField.font = font ?? UIFont.systemFont(ofSize: fontSize)
+            } else {
+                cardField.font = UIFont.systemFont(ofSize: fontSize)
+            }
+        }
+    }
+    
+    func registerFont(_ fontFamily: String) {
+        let controller = UIApplication.shared.delegate?.window??.rootViewController as? FlutterViewController;
+        let bundle = Bundle.main
```
remonh87, line 238: what if the font is located in a separate package and not the main bundle? I do not think this solution facilitates for multi package projects. Either we should fix this or we should name this as a limitation

remonh87 (changes requested):
Have 2 questions / concerns thatt I would like to see answered for the rest LGTM

packages/stripe_ios/ios/Classes/CardFieldView.swift
```diff
@@ -216,13 +216,39 @@ class CardFieldView: NSObject, FlutterPlatformView, STPPaymentCardTextFieldDeleg
...
             }
-            if let fontSize = cardStyle["fontSize"]  as? Int {
-                cardField.font = UIFont.systemFont(ofSize: CGFloat(fontSize))
-            }
+           
             if let placeholderColor = cardStyle["placeholderColor"]  as? String {
                 cardField.placeholderColor = UIColor(hexString: placeholderColor)
             }
             
+            let fontSize = CGFloat(cardStyle["fontSize"]  as? Int  ?? 18);
```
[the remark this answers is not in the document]
jamesblasco: Imply it. 18 is the one given by default by Stripe

packages/stripe_ios/ios/Classes/CardFieldView.swift
jamesblasco: I would prefer to push it like this, or without the registerFont. And then implement it also for packages and also on Android.

remonh87 (approved):

packages/stripe_ios/ios/Classes/CardFieldView.swift
remonh87: let's ship it then :)

PR #235, opened 2021-07-26
Prepare for release 1.2.0

jamesblasco:


remonh87 (approved):

remonh87:
@jamesblasco  check the pub publish errors on the CI:

`* ./CHANGELOG.md doesn't mention current version (1.2.0).` I think at least we should mention latest version?

jamesblasco:
I've published already. It was only a warning and no error. Surprised it exits as an error

remonh87:
I think we should either fix this or remove the check since else the CI will fail for all subsequent PRs.

PR #239, opened 2021-07-27
Handle payment sheet cancel event, by user

SahilSonawaneOfficial:
As the return type presentPaymentSheet method is Future<void>, waiting is done without any approval of payment been succeeded or failed (When the user closes or cancels the payment sheet, in spite, the code beneath was executing which was none other than showing payment success). 

At line 96, I've called retrievePaymentIntent to get the post-payment intent which includes the PaymentIntentsStatus and made a decision for both cases accordingly in a brief.


SahilSonawaneOfficial:
I've got a bit of trouble regarding finding the solution out. Hope this helps others using the example for payment sheet :)

SahilSonawaneOfficial:
I think the issue's already been resolved in #199

PR #244, opened 2021-07-28
Fix StripeException mixin

remonh87:
Make StripeException part of Exception so it is easier to catch it in… one clause


jamesblasco (approved):
Perfect!

Issue #185, opened 2021-07-07
Creating Payment method missing address


hjudoo:
Creating payment method not sending address but  name, email and phone information sent.
 I am using the example no webhooks with cardfield

 On stripe:
Request POST body
{
  "muid": "xxxxxxxx",
  "billing_details": {
    "name": "test30 test",
    "phone": "5555555555",
    "email": "h@gmail.com"
  },

"billing_details": {
          "address": {
            "city": null,
            "country": null,
            "line1": null,
            "line2": null,
            "postal_code": null,
            "state": null
          },

jonasbark:
Which method are you using with this Flutter Stripe plugin where you cannot set an address?

hjudoo:
try {

    final billingDetails = BillingDetails(
      email: 'email@stripe.com',
      phone: '+48888000888',
      address: Address(
        city: 'Quebec',
        country: 'CA',
        line1: '1459  Circle Drive',
        line2: '',
        state: 'quebec',
        postalCode: 'j1x1x1',
      ),
    ); // mocked data for tests

    // 2. Create payment method
    final paymentMethod =
        await Stripe.instance.createPaymentMethod(PaymentMethodParams.card(
      billingDetails: billingDetails,
    ));

jonasbark:
> Creating payment method not sending address but name, email and phone information sent.
what exactly do you mean by that? the address in BillingDetails isn't sent? Isn't sent where? To Stripe? 
Please share more details

hjudoo:
sorry I thought I was clear..my bad. 
I am using the example Payment No Webhooks. The first step in the code is create payment method and sending parameters BillingDetails. The name, email and phone appears in the stripe payment method that was created but not the address and if I look as the post body(in stripe logs) the address was not sent to stripe. The whole process works even if the address is not sent but it was supposed to send the address aswell

jgrody:
This is happening to me as well. 

```
const billingDetails = stripe.BillingDetails(
  email: 'email@stripe.com',
  phone: '+48888000888',
  name: 'Foo Bar',
  address: stripe.Address(
    line1: '123 main st',
    line2: '',
    city: 'San Diego',
    state: 'CA',
    postalCode: '92103',
    country: 'US',
  ),
);
final paymentMethod = await stripe.Stripe.instance.createPaymentMethod(
  stripe.PaymentMethodParams.card(
    billingDetails: billingDetails,
  ),
);
```

This is the output `BillingDetails` object:

```
[...]

PR #284, opened 2021-08-15
stripe_android and stripe_ios: Fix BillingAddress properties being null

musthafa1996:
Fixed `BillingAddress`  properties being null while creating a payment method on Android.
This issue has been addressed at #185.


TesLake (approved):

remonh87 (changes requested):
Thank you for your contribution. I think this pr is valid but should be done in the Stripe react native sdk GitHub repo:  https://github.com/stripe/stripe-react-native .  The issue is if we change it here we would need to manually add our custom additions to the SDK every time we sync with Stripe.  cc: @jonasbark

musthafa1996:
Thanks, @remonh87 for pointing this out. I'll try to fix it there by tomorrow.

PR #308, opened 2021-08-30
Sync - August Update

jamesblasco:
- Add new Card Form
- Add Google Pay button
- Add WeChat example
- Deprecate params from presentPaymentSheet
- Improve Payment Sheet examples
- Refactor ios native side to allow faster sync


remonh87:
> * Add new Card Form
> * Add Google Pay button
> * Add WeChat example
> * Deprecate params from presentPaymentSheet
> * Improve Payment Sheet examples
> * Refactor ios native side to allow faster sync

lets use this as commit message

jonasbark:
Anyone mind adding a changelog?

remonh87 (approved):
In general it looks good I think we need to think of following things:

- [ ]   WeChat and als Google pay should be labeled as Beta. Since there are some issues with google pay in the Stripe react native sdk + Android sdk. For WeChat it is a bit different since I could test it for 50% because I do not have registered the WeChat app .

In case we want to release based on this codebase we should think of:
- [ ]  Will this be the first stable version?
- [ ] Not sure for this pr. Bump versions and do not forget to fix the versions  (`x.x.x` instead of `^x.x.x`.

PR #310, opened 2021-08-30
iOS Apple Pay token

jamesblasco:
Merge after [/examples](https://github.com/flutter-stripe/flutter_stripe/tree/examples) branch

- Adds `createApplePayToken` method that generates a valid Stripe card token from the result of the Apple Pay button from `pay` plugin.
- Adds Apple Pay example with `pay` plugin

### DIFF
--- a/example/ios/Runner.xcodeproj/project.pbxproj
+++ b/example/ios/Runner.xcodeproj/project.pbxproj
@@ -46,6 +46,7 @@
 		97C146FD1CF9000F007C117D /* Assets.xcassets */ = {isa = PBXFileReference; lastKnownFileType = folder.assetcatalog; path = Assets.xcassets; sourceTree = "<group>"; };
 		97C147001CF9000F007C117D /* Base */ = {isa = PBXFileReference; lastKnownFileType = file.storyboard; name = Base; path = Base.lproj/LaunchScreen.storyboard; sourceTree = "<group>"; };
 		97C147021CF9000F007C117D /* Info.plist */ = {isa = PBXFileReference; lastKnownFileType = text.plist.xml; path = Info.plist; sourceTree = "<group>"; };
+		C5F7079826DD2C0800AA3E5B /* Runner.entitlements */ = {isa = PBXFileReference; lastKnownFileType = text.plist.entitlements; path = Runner.entitlements; sourceTree = "<group>"; };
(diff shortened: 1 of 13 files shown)


example/ios/Runner.xcodeproj/project.pbxproj
```diff
@@ -483,17 +486,18 @@
...
-				DEVELOPMENT_TEAM = 8734Y6QF3F;
+				DEVELOPMENT_TEAM = 87RQQPQ85J;
 				ENABLE_BITCODE = NO;
 				INFOPLIST_FILE = Runner/Info.plist;
 				LD_RUNPATH_SEARCH_PATHS = (
 					"$(inherited)",
 					"@executable_path/Frameworks",
 				);
-				PRODUCT_BUNDLE_IDENTIFIER = flutter.stripe.example;
+				PRODUCT_BUNDLE_IDENTIFIER = "flutter.stripe.example-2";
```
remonh87, line 500: why do we need a new bundle identifier we should be able to add Apple Pay later to it right?

packages/stripe/lib/src/stripe.dart
```diff
@@ -117,6 +117,21 @@ class Stripe {
     return isSupported;
   }
 
+  /// Creates a single-use token that represents an Apple Pay credit card’s details.
+  ///
+  /// Tokens are considered legacy, use [PaymentMethod] and [PaymentIntent]
+  /// instead.
+  /// Throws an [StripeError] in case createToken fails.
+  Future<TokenData> createApplePayToken(Map<String, dynamic> payment) async {
```
remonh87, line 125: but in this case token is not legacy right? maybe we should point instead to the stripe Apple Pay method

remonh87 (approved):
this PR also has commits for fix examples. Maybe do a rebase

example/ios/Runner.xcodeproj/project.pbxproj
jamesblasco: I register that bundle identifier with my personal account where I longer don't pay the Developer fee therefore it was not possible to use ApplePay with that identifier

packages/stripe_ios/ios/Classes/StripePlugin+PayPlugin.swift
```diff
@@ -0,0 +1,122 @@
+import Flutter
+import UIKit
+import Stripe
+import PassKit
+
+
+extension StripePlugin {
```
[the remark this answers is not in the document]
jamesblasco: Stripe RN? Yeah maybe it should be a good idea. But it would require a way for them to expose the PKPayment object to the framework layer. On our side, this is now done by the `pay` plugin

PR #423, opened 2021-11-01
Sync React Native 0.2.3

jonasbark:


remonh87 (approved):
should we already bump the version or will that be another pr?

jonasbark:
Yes we should

PR #460, opened 2021-11-22
Add Button Border-Radius for Apple Pay Button

### LINKED ISSUE #448: Add Button Border-Radius for Apple Pay Button
Currently it seems that there is no way to style the build-in Apple Pay Button.

It would be super to set the Border-Radius for the Apple Pay Button, see: https://developer.apple.com/design/human-interface-guidelines/apple-pay/overview/buttons-and-marks/

Thanks for this great package!

Taym95:
fix #448

### DIFF
--- a/packages/stripe/lib/src/widgets/apple_pay_button.dart
+++ b/packages/stripe/lib/src/widgets/apple_pay_button.dart
@@ -17,6 +17,7 @@ class ApplePayButton extends StatelessWidget {
     Key? key,
     this.style = ApplePayButtonStyle.black,
     this.type = ApplePayButtonType.plain,
+    this.cornerRadius = 4.0,
     this.onPressed,
     double? width,
     double? height = _kApplePayButtonDefaultHeight,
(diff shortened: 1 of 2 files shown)


packages/stripe/lib/src/widgets/apple_pay_button.dart
```diff
@@ -17,6 +17,7 @@ class ApplePayButton extends StatelessWidget {
     Key? key,
     this.style = ApplePayButtonStyle.black,
     this.type = ApplePayButtonType.plain,
+    this.cornerRadius = 4.0,
```
remonh87, line 20: personally I would set it to 0.0 since I think it will break the UI for everyone that upgrades the library.

remonh87:
@jamesblasco  can you check this PR?

packages/stripe/lib/src/widgets/apple_pay_button.dart
Taym95: The default value is 4.0 from the doc :

https://developer.apple.com/documentation/passkit/pkpaymentbutton/2999416-cornerradius

jamesblasco (approved):

PR #476, opened 2021-12-08
Add web plugin for beta release

jamesblasco:
Screenshot 2021-12-08 at 17 28 54


example/lib/main.dart
```diff
@@ -22,7 +22,16 @@ class App extends StatelessWidget {
     return DismissFocusOverlay(
       child: MaterialApp(
         theme: exampleAppTheme,
-        home: HomePage(),
```
remonh87, line 25: not for this PR but maybe Navigator2 :)

example/lib/screens/regional_payment_methods/ideal_screen.dart
```diff
@@ -41,7 +42,7 @@ class IdealScreen extends StatelessWidget {
     try {
       await Stripe.instance.confirmPayment(
         clientSecret,
-        PaymentMethodParams.ideal(),
+        PaymentMethodParams.ideal(bankName: kIsWeb ? 'revolut' : null),
```
remonh87, line 45: why hardcoded revolution here?

packages/stripe_checkout/.gitignore
```diff
@@ -0,0 +1,74 @@
+# Miscellaneous
```
remonh87, line 1: I think this file can be removed and should be part of the generic .gitignore file

packages/stripe_checkout/lib/src/platforms/stripe_checkout.dart
```diff
@@ -0,0 +1,126 @@
...
+
+Future<CheckoutResponse> redirectToCheckout({
+  required BuildContext context,
+  required String sessionId,
+  required String publishableKey,
+  String? successUrl,
+  String? canceledUrl,
+}) async {
+  final response = await Navigator.of(context).push(
+    MaterialPageRoute(
```
remonh87, line 20: For now it is fine but should we enforce the material page here? I think in the future maybe people want to inject there own animations. Still for now it should be fine but let's add it to the backlog

packages/stripe_checkout/lib/src/platforms/stripe_checkout.dart
```diff
@@ -0,0 +1,126 @@
...
+    final redirectToCheckoutJs = '''
+var stripe = Stripe(\"$publishableKey\");
+stripe.redirectToCheckout({sessionId: "$sessionId"}).then(function (result) {
+  result.error.message = 'Error'
+});
+123;
+''';
+
+    try {
+      await _webViewController?.runJavascript(redirectToCheckoutJs);
```
remonh87, line 108: I think here we should throw and Exception incase the _webviewController == null. Because now web controller  is null and javascript is executed the redirect won't work

packages/stripe_checkout/pubspec.yaml
```diff
@@ -0,0 +1,58 @@
...
+  js: ^0.6.3
+  freezed_annotation: ^1.0.0
+
+dev_dependencies:
+  flutter_test:
+    sdk: flutter
+  pedantic: ^1.11.0
+  
+
+# For information on the generic Dart part of this file, see the
```
remonh87, line 23: Consider removing this commented section

packages/stripe_checkout/test/stripe_checkout_test.dart
```diff
@@ -0,0 +1,5 @@
+
+
+void main() {
+
```
remonh87, line 4: I think we can make some meaningful tests (e.g. that the redirect is called)

packages/stripe_web/.gitignore
```diff
@@ -0,0 +1,74 @@
+# Miscellaneous
```
remonh87, line 1: same here I think we should have one .gitignore every for the repo

remonh87 (commented):
Wow great work. Just left some first comments. Also have a bit troubles judging the `js` folder. If this is parsed from the Stripe js sdk I do not feel the need of reviewing else I will take a look further

example/lib/main.dart
jamesblasco: True!

example/lib/screens/regional_payment_methods/ideal_screen.dart
jamesblasco: It is required for the web to add a bankName for the ideal payment. I used one of the one available (the only one I have XD)

packages/stripe_checkout/.gitignore
jamesblasco: You can check that there is a .gitignore in every package folder, as it is used by pub publish to ignore files when publishing too.

packages/stripe_checkout/lib/src/platforms/stripe_checkout.dart
jamesblasco: Definitely true!

packages/stripe_checkout/lib/src/platforms/stripe_checkout.dart
jamesblasco: I think this case would never happen, as the onWebView created method should be called before the first onPageFinished is called. Maybe we can use a late variable instead of a nullable one

packages/stripe_checkout/test/stripe_checkout_test.dart
jamesblasco: I don't think this can be done with unit/widget testing as kWeb is constant and WebView are platform views. We should do some integration testing but it is not straightforward right now

packages/stripe_checkout/lib/src/platforms/stripe_checkout_web.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:flutter/widgets.dart';
+import 'checkout.dart';
+import 'stripe_js.dart' as s;
```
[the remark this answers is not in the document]
remonh87: I still see import as s

remonh87:
Looks good to me 2 things that we should do before merging is:

- [ ] Add web support to the main readme file of the package. Maybe also a small disclaimer that it is in beta and experimental?
- [ ] Fix analyser issues (created pr #488 that will fix this already)

remonh87 (approved):

PR #503, opened 2021-12-29
Fix PaymentIntentsStatus parser exception: requires_confirmation is not a payment intent status

solid-software:
Hey, thanks for web implementation! We've tried it out and works well.

However we found and issue with the parser so this PR should fix it.

Issue: PaymentIntentsStatus parser threw an exception requires_confirmation is not a payment intent status
Stripe API (https://stripe.com/docs/api/errors#errors-payment_intent-status)  uses snake_case while PaymentIntentsStatus parser uses PascalCase.

We see that this parsing extension is duplicated - let us know that this is something you want to fix too or there is a reason for having it like this.

Thanks again for your work in web support!


jamesblasco:
Thanks for the PR! Taking an overall look, it seems to make sense the change. I will fully test it tomorrow.

Regarding the duplicate. Setup intent is not implemented yet but yeah it seems that could reuse the PaymentIntentsStatus from the PaymentIntent extension

solid-software:
hey, @jamesblasco, any updates on this?

jamesblasco (approved):

PR #518, opened 2022-01-06
add missing parameters dagerouslyGetFullCardDetails and dangerouslyUp…

### LINKED ISSUE #499: CardField not saving card number, or cvv
`flutter_stripe: ^2.1.0`

CardField isn't updating card number or cvv. Only expiry month and expiry year get updated. The other two remain `null`

Probably because the only arguments it's receiving are:
`{last4: null, expiryMonth: null, expiryYear: null, complete: false, brand: null}`

Even with both of these set to true, number and cvc still aren't saved/returned in the card details.
```
dangerouslyUpdateFullCardDetails: true,
dangerouslyGetFullCardDetails: true,
```

And.. that's because these two parameters are missing from `_MethodChannelCardField` in `lib/widgets/card_field.dart`
```
dangerouslyGetFullCardDetails: widget.dangerouslyGetFullCardDetails,
dangerouslyUpdateFullCardDetails: widget.dangerouslyUpdateFullCardDetails,
```

remonh87:
…dateFullCardDetails to MethodChannelCardfield

Fix #499

### DIFF
--- a/packages/stripe/lib/src/widgets/card_field.dart
+++ b/packages/stripe/lib/src/widgets/card_field.dart
@@ -159,6 +159,8 @@ class _CardFieldState extends State<CardField> {
               style: style,
               placeholder: placeholder,
               enablePostalCode: widget.enablePostalCode,
+              dangerouslyGetFullCardDetails: widget.dangerouslyGetFullCardDetails,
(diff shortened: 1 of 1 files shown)


jonasbark (approved):

jonasbark:
Thanks!

PR #528, opened 2022-01-08
improve Android initialization checks

jonasbark:
This will give better error messages when an Android dependency isn't met


remonh87 (approved):
Awesome this will really help!

PR #551, opened 2022-01-19
[docs]: Theme.AppCompat example link returns 404.

ramanverma2k:
Correct the url link to redirect properly to the file.


jonasbark:
Thanks!

PR #582, opened 2022-02-12
feat: update checkout

jamesblasco:
- Return CheckoutResponse.redirected() when redirectToCheckout is completed successfully on web
 - Remove checkout for web and keep it as a separate package.
 - Require succesfulUrl and canceledUrl when using checkout in a webview


remonh87 (approved):

remonh87:
@jamesblasco  some checks are failing but for the rest it looks good 👍

Issue #609, opened 2022-03-02
HELP NEEDED: Missing return_url parameter for IDEAL payments


ayushin:
This is **not a duplicate** of #225 

While you can set `urlScheme` this does not work with universal links - we really do need to be able to pass `return_url` to IDEAL payment confirmation in order to be able to handle payments across platforms.

Stripe example:
```
    } = await stripe.confirmIdealPayment(clientSecret, {
      payment_method: {
        ideal: elements.getElement(IdealBankElement),
        billing_details: {
          name: 'Jenny Rosen',
        },
      },
      return_url: `${window.location.origin}/ideal?return=true`,
```

ayushin:
Also `window.location.href` should not be used as return_url blindly in `stripe_web.dart` if `return_url` is specified

ayushin:
I could use some help to implement this workaround on Android:

https://github.com/flutter-stripe/flutter_stripe/pull/610

remonh87:
fixed as part of 2.4.0.

davidmigloz:
I see that you can specify a `returnUrl` if you are using the payment sheet:

https://github.com/flutter-stripe/flutter_stripe/blob/4d47412fe49353c955c9ecabe9b1875e4ebc3fbe/packages/stripe_platform_interface/lib/src/models/payment_sheet.freezed.dart#L85

But if you are using the [sdk's webview](https://docs.page/flutter-stripe/flutter_stripe/ideal#using-the-sdks-webview) you don't have any option to specify a `returnUrl`.
So, in iOS we are forced to use custom URL scheme instead of being able to use universal links.

Also for web, `returnUrl` is still hardcoded to `window.location.href`:

https://github.com/flutter-stripe/flutter_stripe/blob/4d47412fe49353c955c9ecabe9b1875e4ebc3fbe/packages/stripe_web/lib/src/web_stripe.dart#L153

Would it be possible to expose `returnUrl` in `confirmPayment()` method so that we can use universal links in iOS and have more flexibility in web?

PR #610, opened 2022-03-02
WIP: return_url in payment intent confirmation

ayushin:
This workaround adds `returnUrl` from `options['returnUrl']` to paymentIntentConfirmation 

Works so far on web & IOS, help needed for android.

#609


ayushin:
@remonh87 Awesome, I am not claiming this to be the right fix and the right place but this works for us now.

Also with android - I am complete lost there - can't even find the `confirmPayment` method so yes, a workaround or at least a pointer appreciated

jonasbark:
This seems to be part of https://github.com/stripe/stripe-react-native/pull/832/files

ayushin:
looking forward to see this properly implemented

PR #660, opened 2022-03-29
Fix the broken link

mshibanami:


jonasbark:
Already handled by another PR - thanks though!

Issue #669, opened 2022-04-07
Payment sheet example does not populate billing details in to Stripe


huckle-panda:
**Describe the bug**

Building the example app and following the payment_sheet_screen.dart flow, the example billing details do not get populated in to the Stripe sandbox when using test card details.

We are seeing similar issue in production, whereby only wallet pay (apple or google) transactions seem to populate customer information reliably.

**To Reproduce**

Build and run example.  Link up Stripe test or production credentials.  Make payment using test card details.   Only customer name seems to appear,  no email or billing address.

**Expected behavior**

The data that is defined in billingDetails should populate in to Stripe.

Thanks

remonh87:
Which version of the library are you using?

remonh87:
yes you are totally right it should be defaultBillingDetails.

remonh87:
Fixed by #680

puneetjoshi04:
Hi, I am still facing this issue with version 3.1.0.

PR #680, opened 2022-04-18
Fix 669 billing details payment sheet

### LINKED ISSUE #285: [PaymentSheet] Allow us to set a default country
When opening the card details panel allow us to add a default country so that payments can be processed faster for local apps.

remonh87:
FIX #285 and #669


packages/stripe_android/android/src/main/kotlin/com/facebook/react/bridge/ReadableMap.java
```diff
@@ -81,7 +81,7 @@ public ReadableType getType(String key) {
                 return ReadableType.Array;
             } else if (value instanceof Number) {
                 return ReadableType.Number;
-            } else if (value instanceof Map) {
+            } else if (value instanceof Map || (value instanceof  JSONObject)) {
```
remonh87, line 84: @jonasbark  this is the most tricky fix. I did some tests and functionality still seems to work

PR #695, opened 2022-04-25
freezed 2.x support

ayushin:


remonh87:
Thanks! you also need to do the same for `stripe_web` because the CI now fails because of that

remonh87 (approved):

Issue #387, opened 2021-10-08
confirmSetupIntent throws error with google pay cardFromToken and valid client secret


klivin:
Hi, thanks for any help with this.

`confirmSetupIntent` fails only in android using google pay token.  Google pay works all the way through with payment intents with same card, just not with setup intents.

**To Reproduce**
```
      // acquire valid payment result from google pay:
      var token = paymentResult['paymentMethodData']['tokenizationData']['token'];
        final tokenJson = Map.castFrom(json.decode(token));
        tokenId = tokenJson['id'];
      }
      final params = PaymentMethodParams.cardFromToken(
          token: tokenId,
          setupFutureUsage: PaymentIntentsFutureUsage.OnSession);

        // Free trial, capture intent for future billing
        final response = await createSetupIntent(customerId);
        final clientSecret = response['clientSecret'];
        // THIS FAILS WITH GOOGLE PAY TOKEN BUT WORKS WITH APPLE PAY TOKEN
        final setupIntentResult = await Stripe.instance.confirmSetupIntent(
          clientSecret,
          params,
      }
```

**Expected behavior**
This should work just like with acquiring an Apple Pay token like we do via:
```
     token = await Stripe.instance.createApplePayToken(paymentResult);
```

The error thrown in Stripe.instance.confirmSetupIntent is **"Card Details not complete"**

The same card works fine in android with confirmPayment when there's an immediate charge, just not as a setup intent (for a free trial for a subscription for instance).  I would use a payment intent, but it won't let me confirm a payment intent with zero charge.

There's no google pay setup intent in the example, and it will fail with same error if I were to change the setup future payment screen to use google pay.

guymclean:
Hi is there any update on this @jamesblasco ?

atearjen:
any update?

daveols:
I've posted a PR just for this fix: https://github.com/flutter-stripe/flutter_stripe/pull/714

If you're impatient you can reference my fork in the meantime, I'll leave it up:

```
  flutter_stripe:
    git:
      url: https://github.com/daveols/flutter_stripe
      ref: android-confirm-setup-intent-hotfix
```

daveols:
I've posted a PR just for this fix: https://github.com/flutter-stripe/flutter_stripe/pull/714

If you're impatient you can reference my fork, I'll leave it up for a few weeks:

```yaml
dependency_overrides:
  stripe_android:
    git:
      url: https://github.com/daveols/flutter_stripe
[...]

PR #714, opened 2022-05-11
feat: add token & paymentMethodId handling to confirmSetupIntent for Cards

daveols:
This is fix for https://github.com/flutter-stripe/flutter_stripe/issues/387 which I've cherry-picked from https://github.com/stripe/stripe-react-native/pull/931.

An a better alternative to merging this PR would be to sync this project with version `0.9.0` of React Native Stripe which was just released and also contains this fix, however there are several breaking changes to accommodate too (I believe we are currently synced with `0.7.0`).

Posting this up anyway for those who may want to reference my fork in the meantime.


remonh87:
Thanks for your contribution but indeed the best way is to sync with the latest version. Syncing with react native v.0.8.0 is a lot of work as the underlying structure of creating payment methods changed totally and we want to make sure nothing breaks.  Hopefully we can sync with v.0.9.0 in the coming days. I close this PR for now but thanks for bringing it to our attention

Issue #718, opened 2022-05-12
Support Flutter 3.0


remonh87:
In #654  there is already a fix but I think we need to do more investigation since Flutter3 implements a new way for the platform bindings.  I think we need to do this

- Upgrade dependencies
- Fix compile issues
- Fix analyzer issues and warnings
- Test all the platformviews (apple pay button, cardfield, cardformfield, aubecsfield and google pay button).

remonh87:
@jonasbark  I can start on this this evening

asaarnak:
Errors after flutter 3 upgrade:
flutter_stripe-2.5.0/lib/src/widgets/card_field.dart:340:22: Warning: Operand of null-aware operation '?.' has type 'WidgetsBinding' which excludes null.
flutter_stripe-2.5.0/lib/src/widgets/card_field.dart:497:26: Warning: Operand of null-aware operation '!' has type 'WidgetsBinding' which excludes null.
flutter_stripe-2.5.0/lib/src/widgets/card_form_field.dart:256:22: Warning: Operand of null-aware operation '?.' has type 'WidgetsBinding' which excludes null.
flutter_stripe-2.5.0/lib/src/widgets/card_form_field.dart:410:26: Warning: Operand of null-aware operation '!' has type 'WidgetsBinding' which excludes null.

vikt0r40:
How to update temporary to 654 fix ?

remonh87:
There is a pr + branch now but it breaks the cardfield on Android. Looks like the changes made in the native platformview has some regression in Flutter 3. We are investigating the issue and will contact the flutter team when necessary. For cardview and cardformview users I do not recommend to upgrade at the moment.

remonh87:
Solved the issue and merged in #721

PR #721, opened 2022-05-12
Support flutter3

remonh87:
Fix for #718 

Let's wait with merging until we got clarity on the cardfield

### DIFF
--- a/packages/stripe_android/android/build.gradle
+++ b/packages/stripe_android/android/build.gradle
@@ -2,14 +2,14 @@ group 'com.flutter.stripe'
 version '1.0-SNAPSHOT'
 
 buildscript {
-    ext.kotlin_version = '1.5.0'
+    ext.kotlin_version = '1.6.21'
     repositories {
         google()
-        jcenter()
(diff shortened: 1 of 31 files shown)


packages/stripe_android/android/build.gradle
```diff
@@ -38,10 +38,10 @@ android {
 
 dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:$kotlin_version"
-    implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.3.1"
+    implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.4.1"
```
jonasbark, line 41: are these dependency updates on purpose? Otherwise, I'd follow the versions from stripe-react-native

remonh87: yeah I thought upgrading those minors couldn't hurt but no problem reverting this. However I would like to upgrade the gradle stuff since it gets rid of a lot of warnings and also the Flutter team recommends these versions now

Issue #764, opened 2022-06-03
Backwards support for flutter 2


DDavidPrime:
Hello @remonh87 .

We are currently using flutter 2 with the latest flutter_stripe version of `2.5.0`. We're however experiencing some issues with it which seem to be addressed in the latest version. However we are not able to use the latest version since we'd need to migrate to flutter 3. Due to some dependencies in our project we are stuck with flutter 2, however.

The issue that we (and most of the flutter community) are experiencing is related to nullable and non-nullable instances of some singletons (such as `WidgetsBinding.instance`,...) which happened between the two different flutter versions. There is however [this solution](https://docs.flutter.dev/development/tools/sdk/release-notes/release-notes-3.0.0#your-code) that could help with providing backwards support (for flutter 2), which I think could prove useful for a lot of developers using this package.

remonh87:
Fixed with release 3.1.0 thanks for the fix!

PR #765, opened 2022-06-03
Backwards support for flutter 2

DDavidPrime:
Provides backwards support for flutter 2 as mentioned in issue number #764

### DIFF
--- a/packages/stripe/lib/src/utils.dart
+++ b/packages/stripe/lib/src/utils.dart
@@ -1,0 +1,1 @@
+T? ambiguate<T>(T? value) => value;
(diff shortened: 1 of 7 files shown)


remonh87:
Thanks! I am fine with this solution. Did you check if AubecsField is using the same widgetsbinding?

remonh87 (approved):

DDavidPrime:
> Thanks! I am fine with this solution. Did you check if AubecsField is using the same widgetsbinding?

Regarding the core stripe project I went through everything and implemented those changes (seems only in four places).
But I just checked and I actually missed the web portion of the project. Will push that in a few minutes.

PR #809, opened 2022-06-26
Provide better guidance in readme

remonh87:
I think we need to refer more to docs.page since we can provide better guidance. Also I really want to get rid of the Cardfield as the main example for handling payments since the `PaymentSheet` is by far the easiest and most smooth integration. Also I want to add subsections for other payment methods plus the wallets (that will be my next doc).

Screenshot 2022-06-26 at 11 52 05


jonasbark:
looks very good to me

Issue #850, opened 2022-07-26
No payment sheet has been initialized yet.


Alfaizkhan:
**main.dart**
```
  WidgetsFlutterBinding.ensureInitialized();
  Stripe.publishableKey = stripePublishableKey;
  Stripe.merchantIdentifier = 'emailID';
  await Stripe.instance.applySettings();
```

 **Initiate Payment**

  ```
Future<void> initPaymentSheet({
    required String? price,
  }) async {
    try {
     await Stripe.instance.initPaymentSheet(
        paymentSheetParameters: SetupPaymentSheetParameters(
          // Main params
          paymentIntentClientSecret: clientSecret,
          customerId: customerId,
          customerEphemeralKeySecret: ephemeralKey,
          // Merchant Name
          merchantDisplayName: 'TEST',
          // Extra params
          applePay: const PaymentSheetApplePay(
            merchantCountryCode: 'US',
          ),
          googlePay: const PaymentSheetGooglePay(
            merchantCountryCode: 'US',
            testEnv: true,
          ),
          style: ThemeMode.system,
        ),
      );
    } catch (e) {
      rethrow;
    }
  }


  /// Display the payment sheet.
  Future<void> confirmPayment({required String price}) async {
    try {
      await initPaymentSheet(price: price);
      // Present Payment sheet to user for payment
      await Stripe.instance.presentPaymentSheet();

      // Book the request.
      bookingRequest();
    } on Exception catch (e) {
      if (e is StripeException) {
        // displaySnackBar(
        //     'Error!', 'Error from Stripe: ${e.error.localizedMessage}');
        log('Error from Stripe: ${e.error.localizedMessage}');
        print('Error from Stripe: ${e.error.localizedMessage}');
        displaySnackBar('Error!', 'Something went wrong!, \nPlease try again.');
      } else {
        log('Error from Stripe: $e');
        displaySnackBar('Error!', 'Something went wrong!, \nPlease try again.');
      }
    }
  }

```
  
 - This **confirmPayment()** is called on Button click.
 - It will wait for Init the payment sheet.
 - After that **presentPaymentSheet** called to present the Payment sheet.

  **In Android, It's working fine.
  In iOS, it's showing error like: No payment sheet has been initialized yet**

jaxnz:
I have the same issue - works in Android.

Not the same as #742

remonh87:
and what if you put a small delay between init and present paymentsheet?

[...]

Issue #385, opened 2021-10-08
StripeException when Presenting Payment Sheet: No payment sheet has been initialized yet


VinceGee:
**Describe the bug**
I am getting the following error when trying to implement flutter_stripe in my app
flutter: StripeException(error: LocalizedErrorMessage(code: FailureCode.Failed, localizedMessage: No payment sheet has been initialized yet, message: No payment sheet has been initialized yet, stripeErrorCode: null, declineCode: null, type: null))

**To Reproduce**
1. I am getting my paymentIntent from the server, using the following:
```
MyResponse myResponse = await CheckoutController.processThePayment(orderId);
    if (myResponse.success) {
      paymentIntentData = myResponse.data;

    } else {
      ApiUtil.checkRedirectNavigation(context, myResponse.responseCode);
      showMessage(message: myResponse.errorText);
    }
```
2. Initializing the paymentSheet as follows:
```
await Stripe.instance.initPaymentSheet(
        paymentSheetParameters: SetupPaymentSheetParameters(
          paymentIntentClientSecret: paymentIntentData['paymentIntent'],
          applePay: true,
          googlePay: true,
          style: ThemeMode.dark,
          merchantCountryCode: 'UK',
          merchantDisplayName: 'VinceShop'
        ));

    setState(() {});
    print('initialised');
    displaySheet();
```
3. The displaySheet() is defined as follows:
```
Future<void> displaySheet() async {
    try {
      print('trying to present');<-----------This is printing
      await Stripe.instance.presentPaymentSheet();
      print('not reaching here?');<-----------This is NOT printing

      setState(() {
        showMessage(message: "Order paid successfully.");
        paymentIntentData = null;
      });
    } catch (e){
      print(e);
      showMessage(message: "Failed to make the payment.");
    }
  }
```
4. The StripeKey is initialized in the main():
```
    Stripe.publishableKey = pubKey;
    Stripe.merchantIdentifier = 'VinceShop';
    await Stripe.instance.applySettings();
```

**Expected behavior**
Show PaymentSheet, where client can enter their payment details to confirm and complete a transaction

**Smartphone / tablet**
 - Device: [iPhone 12 Pro Max Emulator and Huawei G8]
 - OS: [e.g. iOS 14.2, Android 6.0.1]
 - Package version: [e.g. 1.3.0]
 - Flutter version [e.g. Flutter 2.5.1 • channel stable.]

**Additional context**
[...]

PR #860, opened 2022-07-29
fix: No payment sheet has been initialized yet on iOS

jaxnz:
Assertion to make sure merchantIdentifier is set if using apple pay. 
At the moment, it just errors saying "No payment sheet has been initialized yet"

#850 & #385


jonasbark:
Thank you!

PR #870, opened 2022-08-12
Changed com.stripe:stripe-android version

Uvais-Mohammad:
Changed com.stripe:stripe-android version from 17.1.1  to 20.8.0 (latest version of Stripe Android Sdk)

### DIFF
--- a/docs/sheet.mdx
+++ b/docs/sheet.mdx
@@ -205,7 +205,7 @@ build.gradle
 
 ```
 dependencies {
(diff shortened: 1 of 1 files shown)


remonh87:
@Uvais-Mohammad  thank you for contributing to our repository however we need to keep in sync with the stripe sdk of the react native library as we are using these bindings to communicate contents through the platform channels.  Your pr has the potential of breaking the library. 

For sure the sdk will updated with the next sync.

PR #939, opened 2022-09-21
Fix: implemented handleURLCallback method for iOS iDeal payment.

saiful-apx:
`handleURLCallback` method was needed for iOS iDeal payment to close the safari in-app webview. Currently, it happens from `swift` side which doesn't provide the `returnUrl` deep-link in flutter side. So we can not navigate the user from the deep link to the iDeal payment success screen.


jonasbark:
great catch, thank you

remonh87 (approved):
thank!! looks good to me will take it along with the next sync

PR #954, opened 2022-10-03
provide correct payment method options

### LINKED ISSUE #911: setting PaymentIntentsFutureUsage in PaymentMethodOptions is not saving card
**Describe the bug**
I have set up the paymentMethod options as follows: 

`PaymentMethodOptions saveOptions = const PaymentMethodOptions(
      setupFutureUsage: PaymentIntentsFutureUsage.OnSession,
    );`

and use this option when confirming a payment via: 

`paymentIntent = await Stripe.instance.confirmPayment(
          secret,
          PaymentMethodParams.card(
            options: saveOptions,
            paymentMethodData: PaymentMethodData(
              billingDetails: billing,
            ),
          ),
        );`

however when looking at the payment in Stripe console, although the payment succeeds, the "setup_future_usage" parameter is never set in the `/v1/payment_intents/pi_.../confirm` API call, and so the card details are never saved. I can see that this is successfully set to "setup_future_usage": "on_session", when paying via our working web integration.

**To Reproduce**
1. we create paymentIntent via our api, which returns secret/publishable_key.
2. set publishable_key/merchantIdentifier on Stripe.instace.
3. attempt to confrim payment using Stripe.instance.confirmPayment() as above.
4. payment succeeds but does not save card details for future use.

**Expected behavior**
"setup_future_usage" to be set and card to appear in customer's Payment Methods 

**Smartphone / tablet**
 - Device: ios simulator (iphone11 / ios 15.5) & iPhone 11 (ios 15.6.1)
 - OS:ios 15.5
 - Package version: 5.0.0
 - Flutter version 3.0.5

**Additional context**
Hopefully I have configured something wrongly, although I was able to save cards previously with this method (maybe!) 🤔

remonh87:
Fix #911  unfortunately this is a breaking change but it couldn't be avoided. Now the dart sdk is in line with react native sdk


remonh87:
Fix #911

PR #1052, opened 2022-12-02
feat: stripe js

jamesblasco:
Adds a new package with the dart bindings for the stripe.js. The package implements the most used endpoints but the whole api is not fully supported yet. 

In addition it allows to load the stripe.js library using the `loadStripe()` method


packages/stripe_js/README.md
```diff
@@ -0,0 +1,3 @@
+# Stripe.js bindings for dart2js
+
+This package is in a very early stage so be cautious to use it in production.
```
jonasbark, line 3: impressive!
can you add a bit of information on how the bindings were generated, as in with what tool + commands?

remonh87 (approved):
Amazing work Jaime 👏  

reviewing everything was a bit to much but I did check the changes in the shared packages and let's see the feedback from the community and gradually improve

PR #1162, opened 2023-03-12
fix created type for token

jonasbark:
it's not a string - it's an integer


remonh87 (approved):

jaredanderton:
Is there an ETA of when this will be released for use with [flutter_stripe](https://pub.dev/packages/flutter_stripe)?

Issue #813, opened 2022-06-28
CardField layout bug on Android 9(28)


AAkira:
**Describe the bug**

`CardField` not working on Android9.

![screenshot-220628172912](https://user-images.githubusercontent.com/3386962/176132340-b0d606ec-2d15-4679-9910-075328ab67d3.png)

**To Reproduce**

1. Launch `ThemeCardExample`

**Smartphone / tablet**
 - Device: Pixel2
 - OS: Android 9
 - Package version: 3.1.0(curernt master)
 - Flutter version: 2.10.0 

**Additional context**

It is also reproduced in the emulator.

* `CardFormField` has the same bug

![screenshot-220628173654](https://user-images.githubusercontent.com/3386962/176134086-b3b4e1da-3711-4240-8b3b-ff19f79b8928.png)

oboenikui:
Related flutter issues:

- https://github.com/flutter/flutter/issues/106543
- https://github.com/flutter/flutter/issues/104686

This bug seems to be caused by a destructive change in Flutter 3.
Before API 29, transparent platform views do not work correctly.

It seems to be fixed if using initExpensiveAndroidView instead of initSurfaceAndroidView here.
(Please note that this method is only available in Flutter 3 or later)
https://github.com/flutter-stripe/flutter_stripe/blob/main/packages/stripe/lib/src/widgets/card_field.dart#L580

Another workaround is to allow the background color to be set from the flutter side in the views; in CardField it is fixed to a transparent color and cannot be set.

remonh87:
Thanks @oboenikui  for the research. 

According to the docs it makes sense to use the expensive version: 

```
When this factory is used, the Android view and Flutter widgets are composed at the
Android view hierarchy level.
This is only useful if the view is a Android SurfaceView. However, using this method
has a performance cost on devices that run below 10, or underpowered devices.
In most situations, you should use [initAndroidView].
```

However this forces us to abandon version 2 and has more known issues: https://github.com/flutter/flutter/issues/107313 . I will keep an eye on it. btw for people that want to try it out: use branch `poc-use-expensive-androidview`

ethancadoo:
I am also having this issue on the latest version of the plugin (5.0.0). This really makes the Android implementation unusable without a for to apply the "expensive-androidview" fix. Please consider merging that fix into main. Thank you!

remonh87:
@ethancadoo  can you confirm that the `poc-use-expensive-androidview` branch works for you?

[...]

PR #1165, opened 2023-03-13
Update platform view for card element

garlicPasta:
Hi,
our app suffered from really bad performance when the `CardField` was used (60fps dropped to 14 fps as soon as the Card field was shown). Scrolling the view became very laggy.

While looking into it i found that `PlatformViewsService.initExpensiveAndroidView` is called. I changed to `initAndroidView` and the performance got a lot better (+10fps) and the functionality was still working. I am not really sure, if that makes problems in certain scenarios, but according to the documentation, androids picks the platform best supported platform view.

If u guys think it might become a problem, i can add like a `flag` or something similar, so the developer can choose between the two modes.

Regards

### DIFF
--- a/packages/stripe/lib/src/widgets/card_field.dart
+++ b/packages/stripe/lib/src/widgets/card_field.dart
@@ -602,7 +602,7 @@ class _AndroidCardField extends StatelessWidget {
       ),
       onCreatePlatformView: (params) {
         onPlatformViewCreated(params.id);
-        return PlatformViewsService.initExpensiveAndroidView(
+        return PlatformViewsService.initAndroidView(
           id: params.id,
           viewType: viewType,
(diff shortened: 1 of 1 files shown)


remonh87:
yes you are right the cardfield is not very performant. However we need to use the expensive android surface view to prevent rendering issues like: https://github.com/flutter-stripe/flutter_stripe/issues/813 . 

Personally I wouldn't recommend using the payment sheet as this provides by far the most smooth experience. Nowadays you can even style the payment sheet.

jonasbark:
As you've mentioned a viable approach could be the addition of a flag

remonh87:
Thanks I have incorporated the flag and it will be part of the next version of this library. Closing this PR.

PR #1172, opened 2023-03-17
Add support of Bacs Debit payment method to confirm SetupIntent

alexshikov:
Add missing data Flutter types and conversation to/from native platform types when working with UK Bank Account payment methods (Bacs Direct Debit)


jonasbark:
Thanks for the PR!

In general, we do not want to deviate from the stripe-react-native original sources as we are syncing their source code with ours. It would be best to create an issue (or even the PR) to their repository:
https://github.com/stripe/stripe-react-native

alexshikov:
@jonasbark thank you for the feedback. Does that mean that this PR won't be accepted until a similar one get into the react-native repository?
And by _syncing_ you mean the code is somehow auto-generated from react-native to flutter?

remonh87:
@alexshikov  sorry for the late reaction I think this slipped a bit between other work. Thanks for the PR but indeed we need to fix this in the react native sdk first. Reason is that else our native iOS and Android wrappers will be different than the offical stripe sdk and will make syncing difficult. I would recommend filing an issue there, I can make sure it will come under their attention. Will close it for now.

PR #1229, opened 2023-05-08
sync with react native 0.27.1

jonasbark:


remonh87 (approved):

jonasbark:
Android doesn't seem to compile - I'll check tomorrow

remonh87:
> Android doesn't seem to compile - I'll check tomorrow

I pushed a fix in last commit . I checked the react native interface and they also cast it to a double instead of a float.

PR #1231, opened 2023-05-08
Fix ApplePayButton styling

Kuodster:
Fix ApplePayButton passing an incorrect key for the styling option to the platform channel


remonh87 (approved):
thanks!

Issue #1246, opened 2023-05-19
Calling the confirmPayment method without a params object causes exception


tbuczkowski:
**Describe the bug**
I have a `paymentIntentClientSecret` that is created on the back-end with a payment method already attached. From my understanding, I should be able to call the `confirmPayment` method without providing the `params` object in a scenario like mine. Here's the method call in my code:
```
await _stripe.confirmPayment(paymentIntentClientSecret: clientSecret);
```
which results in the following errors:
IOS:
```
flutter: PlatformException(Invalid Params, , , null)
flutter: 
#0      JSONMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:171:7)
#1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:310:18)
<asynchronous suspension>
#2      MethodChannel.invokeMapMethod (package:flutter/src/services/platform_channel.dart:510:43)
<asynchronous suspension>
#3      MethodChannelStripe.confirmPayment (package:stripe_platform_interface/src/method_channel_stripe.dart:95:20)
<asynchronous suspension>
#4      Stripe.confirmPayment (package:flutter_stripe/src/stripe.dart:410:29)
<asynchronous suspension>
```
Android:
```
I/flutter ( 9138): PlatformException(error, org.json.JSONObject$1 cannot be cast to org.json.JSONObject, null, java.lang.ClassCastException: org.json.JSONObject$1 cannot be cast to org.json.JSONObject
I/flutter ( 9138): 	at com.flutter.stripe.StripeAndroidPlugin.onMethodCall(StripeAndroidPlugin.kt:287)
I/flutter ( 9138): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:258)
I/flutter ( 9138): 	at io.flutter.embedding.engine.dart.DartMessenger.invokeHandler(DartMessenger.java:295)
I/flutter ( 9138): 	at io.flutter.embedding.engine.dart.DartMessenger.lambda$dispatchMessageToQueue$0$io-flutter-embedding-engine-dart-DartMessenger(DartMessenger.java:322)
I/flutter ( 9138): 	at io.flutter.embedding.engine.dart.DartMessenger$$ExternalSyntheticLambda0.run(Unknown Source:12)
I/flutter ( 9138): 	at android.os.Handler.handleCallback(Handler.java:938)
[...]

PR #1286, opened 2023-06-21
FIX: #1246 handle empty params on confirmPayment

tbuczkowski:
It's been over a month since reporting the #1246 issue, and I'm running out of time to deliver a feature dependant on this fix, so I decided to have a go at trying to fix it myself. The details of why the `confirmPayment` method should handle null this way, are described in my issue description. I have next to none prior experience with Swift or Kotlin, so please double check that my changes don't break any other use cases.


remonh87 (approved):
Sorry it was an extremely busy months. Thanks for picking up the issue. @jonasbark  can you double check the android side?

jonasbark (approved):

PR #1294, opened 2023-06-29
feat: add klarna for web

jamesblasco:
- Adds klarna for web
- Implements retrievePaymentIntent for web


example/lib/main.dart
```diff
@@ -8,7 +8,6 @@ void main() async {
   WidgetsFlutterBinding.ensureInitialized();
   Stripe.publishableKey = stripePublishableKey;
   Stripe.merchantIdentifier = 'merchant.flutter.stripe.test';
-  Stripe.urlScheme = 'flutterstripe';
```
jonasbark, line 11: done on purpose?

jamesblasco: Yes, url scheme should match a custom scheme. On web it is used to redirect back after a payment has been successful and defaults to the window url

remonh87:
close since it is outdated

PR #1308, opened 2023-07-11
Updating values to the new API

Pablo-Aldana:
PaymentInet and ephemeralKey are not longer used  and makes the paymentSheet not been initialised .  it has been changed for the new      "client_secret" as per the API reference.

### DIFF
--- a/example/lib/screens/payment_sheet/payment_sheet_screen.dart
+++ b/example/lib/screens/payment_sheet/payment_sheet_screen.dart
@@ -88,11 +88,11 @@ class _PaymentSheetScreenState extends State<PaymentSheetScreen> {
       await Stripe.instance.initPaymentSheet(
         paymentSheetParameters: SetupPaymentSheetParameters(
           // Main params
-          paymentIntentClientSecret: data['paymentIntent'],
+           paymentIntentClientSecret: data['client_secret'],
+
           merchantDisplayName: 'Flutter Stripe Store Demo',
           // Customer params
           customerId: data['customer'],
-          customerEphemeralKeySecret: data['ephemeralKey'],
           // Extra params
           primaryButtonLabel: 'Pay now',
           applePay: PaymentSheetApplePay(


remonh87:
Thanks for the pr but our app is not contacting directly the stripe api (this is a bad practice and our request could be manipulated.  We use a [example backend](https://github.com/flutter-stripe/flutter_stripe/tree/main/example/server) that comes from Stripe.  Also about the ephermal key. I think it is still used as it is in the docs. https://stripe.com/docs/payments/accept-a-payment?platform=android

Pablo-Aldana:
Hi Remonh 

Thank you very much for taking the time  on reviewing my PR. It is quite weird tho, I'm also using custom backend and not calling directly to Stripe, and the example here is to the PaymentIntent API  so that the paymentsheet can be initialised. I have double checked and when I called PaymentIntent, Stripe return the following JSON

```
{
    "id": "pi_3NSMllAECsC2I7K30Q3qBZVZ",
    "object": "payment_intent",
    "amount": 1000,
    "amount_capturable": 0,
    "amount_details": {
        "tip": {}
    },
    "amount_received": 0,
    "application": null,
    "application_fee_amount": 123,
    "automatic_payment_methods": {
        "allow_redirects": "always",
        "enabled": true
    },
    "canceled_at": null,
    "cancellation_reason": null,
    "capture_method": "automatic",
    "client_secret": "pi_3NSMllAECsC2I7K30Q3qBZVZ_secret_r2tVixqjL2i7GcIre5geLSX2P",
    "confirmation_method": "automatic",
    "created": 1689005821,
    "currency": "eur",
    "customer": null,
    "description": null,
    "invoice": null,
    "last_payment_error": null,
    "latest_charge": null,
    "livemode": false,
    "metadata": {},
    "next_action": null,
    "on_behalf_of": null,
    "payment_method": null,
    "payment_method_options": {
        "bancontact": {
            "preferred_language": "en"
        },
        "card": {
            "installments": null,
            "mandate_options": null,
            "network": null,
            "request_three_d_secure": "automatic"
        },
        "eps": {},
        "giropay": {},
        "ideal": {},
        "link": {
            "persistent_token": null
        }
    },
    "payment_method_types": [
        "card",
        "bancontact",
        "eps",
        "giropay",
        "ideal",
        "link"
    ],
    "processing": null,
    "receipt_email": null,
    "review": null,
    "setup_future_usage": null,
    "shipping": null,
    "source": null,
    "statement_descriptor": null,
    "statement_descriptor_suffix": null,
    "status": "requires_payment_method",
    "transfer_data": {
        "destination": "acct_1NRGhmPKlISh9SHN"
    },
    "transfer_group": null
}
```

In this JSON there is not ephemeralKey neither paymentIntent just client_secret and without these changes I could not make work the payment sheet.

Just in case I checked the Stripe API: https://stripe.com/docs/api/payment_intents and it does not provide ephemeral or paymentIntent value either.

According to your link it would be the step number 3 when it is called to  paymentIntents.create
```
 const paymentIntent = await stripe.paymentIntents.create({
    amount: 1099,
    currency: 'eur',
    customer: customer.id,
    automatic_payment_methods: {
      enabled: true,
    },
  });
```
This call from my understanding does not provide the values used later on Step 4:

```
 paymentIntentClientSecret = responseJson.getString("paymentIntent")
        customerConfig = PaymentSheet.CustomerConfiguration(
          responseJson.getString("customer"),
          responseJson.getString("ephemeralKey")
        )
```

am I understanding something incorrectly?

Thanks a lot!

jonasbark:
Feel free to create a new Q&A discussion topic for this 👍

PR #1309, opened 2023-07-11
Squashed MediaQuery InheritedModel

furkankurt:
When some aspect of the flutter `MediaQuery` changes, every widget which depends on `MediaQuery.of(context)` is rebuilt. Most of these rebuilds are unnecessary, since most changes to `MediaQuery` do not change more than 1 or 2 fields at a time (window resizing, system brightness changing, system font scale changing).

MediaQuery does not change often on mobile, but there is one place that `MediaQuery` is rapidly updated on mobile: on **iOS**, as the keyboard opens, `MediaQuery` is updated every frame with the only change being the keyboard insets.

This change makes `MediaQuery` an _InheritedModel_ rather than an _InheritedWidget_, so any widget which knows it only depends on a specific property of `MediaQuery` the ability to declare that when reading the `MediaQuery` from the _context_.

For example,
```dart
MediaQuery.of(context).viewInsets
```
could become
```dart
MediaQuery.viewInsetsOf(context)
```
as a performance optimisation.


remonh87:
First thanks for the PR! I like the change but am not super convinced about this feature yet since it will force all our users to upgrade to version 3.10. @jonasbark  @jamesblasco  what do you think?

jonasbark:
I agree. It should not be merged, yet.

remonh87:
@jonasbark  I feel we can integrate this now what do you think?

cedvdb:
semver should be upgraded if this is a breaking change

furkankurt:
> semver should be upgraded if this is a breaking change

I don't think that's a breaking change @cedvdb 
In the [official Flutter documentation](https://api.flutter.dev/flutter/widgets/MediaQuery/viewInsetsOf.html), they say: 

> Prefer using this function over getting the attribute directly from the [MediaQueryData](https://api.flutter.dev/flutter/widgets/MediaQueryData-class.html) returned from [of](https://api.flutter.dev/flutter/widgets/MediaQuery/of.html), because using this function will only rebuild the context when this specific attribute changes, not when any attribute changes.

cedvdb:
@furkankurt I based my comment on the comment chain above :

> First thanks for the PR! I like the change but am not super convinced about this feature yet since it will force all our users to upgrade to version 3.10

remonh87:
changes are done in another mr.

Issue #1321, opened 2023-07-19
After confirm payment it throw a exception


EliaTolin:
**Describe the bug**
After complete payment stripe throw a Exception

**To Reproduce**
1. Entering number 4242 4242 4242 4242 into the cardfield.
2. Tapping the confirm button.
3. Observe a failure with exception (including the part of the stack trace, belonging to this package) …

**Smartphone / tablet**
 - Device: iPhone14 Simulator
 - OS: 16.4
 - Package version: 9.3.0
 
  Flutter 3.10.6 • channel stable • https://github.com/flutter/flutter.git
  Framework • revision f468f3366c (7 giorni fa) • 2023-07-12 15:19:05 -0700
  Engine • revision cdbeda788a
  Tools • Dart 3.0.6 • DevTools 2.23.1

**Additional context**
Code: 
```
 Future<void> initStripe() async {
    if (paymentDataKey != null) {
      await Stripe.instance.initPaymentSheet(
        paymentSheetParameters: SetupPaymentSheetParameters(
          // Main params
          merchantDisplayName: PaymentConstants.merchantDisplayName,
          paymentIntentClientSecret: paymentDataKey?.clientSecret,
        ),
      );
    } else {
      throw Exception();
    }
  }

  Future<bool> openPaymentSheet() async {
    try {
      await Stripe.instance.presentPaymentSheet();
      return true;
    } catch (error, stacktrace) {
      if (error is StripeException) {
        if (error.error.code == FailureCode.Canceled) {
          return false;
        }
      }
      log(error.toString());
      log(stacktrace.toString());
      throw Exception();
    }
  }
```

Exception: 
`type 'List<dynamic>' is not a subtype of type 'Map<String, dynamic>?'`

Stacktrace: 
```
MethodChannelStripe.presentPaymentSheet (package:stripe_platform_interface/src/method_channel_stripe.dart:263:37)
      <asynchronous suspension>
      #1      Stripe.presentPaymentSheet (package:flutter_stripe/src/stripe.dart:517:12)
      <asynchronous suspension>
      #2      SlotPaymentController.openPaymentSheet (package: foo/controller/payment/slot_payment_controller.dart:40:7)
      <asynchronous suspension>
      #3      _SlotPaymentPageState.openPayment (package:foo/pages/payment/slot_payment_page.dart:164:11)
      <asynchronous suspension>

```

jorger5:
[...]

PR #1325, opened 2023-07-20
fix: convert from array to dictionary

jorger5:
This fixes https://github.com/flutter-stripe/flutter_stripe/issues/1321 where Flutter would expect a `Map<String,dynamic>?` while iOS was sending an empty array.

This resulted in  `type 'List<dynamic>' is not a subtype of type 'Map<String, dynamic>?'` exception.

### DIFF
--- a/packages/stripe_ios/ios/Classes/Stripe Sdk/StripeSdk.swift
+++ b/packages/stripe_ios/ios/Classes/Stripe Sdk/StripeSdk.swift
@@ -197,7 +197,7 @@ class StripeSdk: RCTEventEmitter, STPBankSelectionViewControllerDelegate, UIAdap
                     paymentSheetViewController = nil
                     switch paymentResult {
                     case .completed:
-                        resolve([])
+                        resolve([:])
                         self.paymentSheet = nil
                     case .canceled:
                         resolve(Errors.createError(ErrorType.Canceled, "The payment has been canceled"))


EliaTolin (approved):
It's good!

EliaTolin:
> We are syncing the file from here: https://github.com/stripe/stripe-react-native/blob/master/ios/StripeSdk.swift#L207
> 
> So, to ensure that the fix isn't overwritten each sync, we'd need to adjust one layer above: https://github.com/flutter-stripe/flutter_stripe/blob/main/packages/stripe_ios/ios/Classes/StripePlugin.swift#L198

Do you suggest then to transform the result to map above?

EliaTolin:
@jonasbark I think it is very important to fix the problem, Stripe does not currently work on iOS. See the issue #1321

jonasbark:
> > We are syncing the file from here: https://github.com/stripe/stripe-react-native/blob/master/ios/StripeSdk.swift#L207
> > So, to ensure that the fix isn't overwritten each sync, we'd need to adjust one layer above: https://github.com/flutter-stripe/flutter_stripe/blob/main/packages/stripe_ios/ios/Classes/StripePlugin.swift#L198
> 
> Do you suggest then to transform the result to map above?

Yes - as in leave the source code as it is, but check for the result in the layer above. E.g. if result == [] => return [:]

EliaTolin:
Move to #1333.

jorger5:
Closing as it was moved to #1333

PR #1335, opened 2023-07-25
Sync with React Native 0.29.0

jonasbark:


packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/PaymentSheetFragment.kt
```diff
@@ -120,6 +127,34 @@ class PaymentSheetFragment(
...
+          cause = Exception("Tried to call confirmHandler, but no callback was found. Please file an issue: https://github.com/stripe/stripe-react-native/issues"),
+          displayMessage = "An unexpected error occurred"
+          )
+      }
+      val params = Arguments.createMap().apply {
+        putMap("paymentMethod", mapFromPaymentMethod(paymentMethod))
+        putBoolean("shouldSavePaymentMethod", shouldSavePaymentMethod)
+      }
+
+      stripeSdkModule.sendEvent(context, "onConfirmHandlerCallback", params)
```
jonasbark, line 143: @remonh87 fyi this event now seems to be supported on Android as well

remonh87 (approved):

Issue #1211, opened 2023-04-19
PayPal payments do not work


creativecreatorormaybenot:
#1194 was closed as PayPal support should have been added in version `9.1.1` (see #1171).

However, I still see the same behavior. When trying to confirm a payment intent using `PaymentMethodParams.payPal`, nothing happens. There is no error message and the future completes as a noop (no web view is opened, nothing).

---

The implementation looks as follows:

```dart
final intent = await _createPaymentIntent(ref, mode);

Stripe.publishableKey = intent.publishableKey;
const params = PaymentMethodParams.payPal(
  paymentMethodData: PaymentMethodData(
    billingDetails: BillingDetails(
      email: ..,
      address: Address(
        ...,
      ),
      phone: ..,
      name: ..,
    ),
    shippingDetails: ShippingDetails(
      address: Address(
        ...,
      ),
      phone: ..,
      name: ..,
      carrier: ..,
      trackingNumber: ..,
    ),
  ),
);

await stripe.confirmPayment(
  paymentIntentClientSecret: intent.clientSecret,
  data: params,
);
```

In our default implementation, we do not even pass `billingDetails` or `shippingDetails` and this works for credit card, Apple Pay, and Google Pay. However, I figured that this may be an issue and therefore added all possible parameters with valid values.

Still, this does not change anything. On the backend payment intent side, I have chatted with the Stripe team and it all looked good. Do you think this is an issue with the `flutter_stripe` implementation or should I contact Stripe again?

creativecreatorormaybenot:
cc @jamesblasco @jonasbark @remonh87

jonasbark:
Works fine on my side with the PayPalScreen from the example app.

jonasbark:
Closing issue due to lack of response

nicklbaert:
@jonasbark Sorry for the delay. We needed to pause working on this due to an internal emergency. Contrary to what the example app is doing, we are creating a subscriptionIntent instead of a paymentIntent. 

This is the output from stripe.confirmPayment(..)

`StripeException(error: LocalizedErrorMessage(code: FailureCode.Failed, localizedMessage: There was an unexpected error -- try again in a few seconds, message: Missing required param: mandate_data[customer_acceptance][online]., stripeErrorCode: parameter_missing, declineCode: null, type: invalid_request_error))`

I passed this as the data parameter:
`const params = PaymentMethodParams.payPal(
    paymentMethodData: PaymentMethodData(
[...]

PR #1341, opened 2023-07-26
Fix PayPal payments

JonasWanke:
Fixes: #1211

The serialization of `MandateDataCustomerAcceptance` was incorrect: In Stripe's API, it corresponds to [a sealed union representing an `online` or an `offline` mandate](https://stripe.com/docs/api/setup_intents/confirm#confirm_setup_intent-mandate_data). In Stripe's Android SDK, it's [modelled the same way, though only `online` is supported](https://github.com/stripe/stripe-android/blob/1be82b108cba58e55893aea75fa7da4d5c640a68/payments-core/src/main/java/com/stripe/android/model/MandateDataParams.kt). To maintain backwards compatibility, I kept the existing class structure and only added custom `fromJson(…)` and `toJson(…)` implementations, though a cleaner design would probably be a named constructor `MandateDataCustomerAcceptance.online(…)`.

The other problem was that `PaymentMethodType` was missing a PayPal variant, meaning the native SDK's result message couldn't be parsed.

Now, payment and setup intents for subscriptions in Stripe are working for us both on Android and iOS with the following parameters:

```dart
const params = PaymentMethodParams.payPal(
  paymentMethodData: PaymentMethodData(
    mandateData: MandateData(
      customerAcceptance: MandateDataCustomerAcceptance(
        ipAddress: MandateDataOnlineData(
          ipAddress: '127.0.0.1',
          userAgent: 'device',
        ),
      ),
    ),
  ),
);
```


remonh87:
@JonasWanke  thanks for the pr. Can you run `dart format` in your pr so we do not see the formatting changes in your pr?

JonasWanke:
Hi @remonh87, I just ran `dart format .`, but it only changed another file. The first commit I pushed (https://github.com/flutter-stripe/flutter_stripe/pull/1341/commits/495a7fa3f5bb7f89074d3ca4db766a49fc84c101) was the result of running build_runner, which already changed the files. I'm guessing that either some formatting behavior changed or they were otherwise formatted differently before, since both the build_runner and `dart format` agree on the current formatting

remonh87 (approved):
thanks

bonnyfone:
thanks for the fix! any timeline for when it will be published on pub.dev?

artemsorochan:
Any timeline on publishing a new version with this MR would be appreciated. Thanks in advance!

artemsorochan:
Any update on this?

PR #1415, opened 2023-09-15
android_stripe class names changes

TetrixGauss:
Change the android_stripe class names to avoid the mutliple definitions of all the classes when the stripe package is used with jitsi flutter packages


TetrixGauss:
if you accept the request we need to change the pubspec paths for stripe_android

TetrixGauss:
@remonh87
@jamesblasco
@jonasbark

jonasbark:
Thanks for that PR.

This solution would be a one-time fix. As we are syncing the sources from https://github.com/stripe/stripe-react-native we would need to apply this every single time, which is not what we would want.

So two solutions:
1. adjust jitsi_meet instead
2. have an automated way of renaming the java package on-demand, e.g. using a script

Issue #1423, opened 2023-09-20
Phone field of ShippingDetails is not being sent in the payload to Stripe


h3nr1ke:
**Describe the bug**
The module is not sending the phone information in the payload when using Affirm payment method.
It was tested using Android and iOS 

**To Reproduce**

1 - Create a `BillingDetails` and `ShippingDetails` 

```dart
    BillingDetails billingDetails = BillingDetails(
      email: "userAffirm@email.com",
      name: "User Name Billing",
      address: "Any Valid Address",
      phone: "9888888888",
    );

    ShippingDetails shippingDetails = ShippingDetails(
      name: "User Name Shipping",
      address: "Any Valid Address",
      phone: "9999999999",
    );

```

2. Use the above object to create a `PaymentMethodParams` using the `PaymentMethodParams.affirm` factory

```dart
    // generate the paymentMethodParams
    PaymentMethodParams params = PaymentMethodParams.affirm(
      paymentMethodData: PaymentMethodData(
        billingDetails: billingDetails,
        shippingDetails: shippingDetails,
      ),
    );
```

3. Call the `confirmPayment` method

```dart
      PaymentIntent confirmation = await Stripe.instance.confirmPayment(
        paymentIntentClientSecret: "YOUR_SECRET_HERE,
        data: params,
      );

```

**Expected behavior**
Must to see the Phone Number field under the Shipping area in the stripe website payment details page

**Smartphone / tablet**
 - Device: Any Android and iPhone
 - OS: Android and iOS
 - Package version: 9.4.0
 - Flutter version - 3

**Additional context**

It works ok for `BillingDetails`.

Order done with the current plugin

![image](https://github.com/flutter-stripe/flutter_stripe/assets/349338/47c41847-3e52-4c5c-9e0c-3e4f70e874e4)

Payload without the phone field

![image](https://github.com/flutter-stripe/flutter_stripe/assets/349338/c5ee42c9-94dd-4cbe-8e09-20a22e352f7f)

Payment created with a modified version sending the phone field

![image](https://github.com/flutter-stripe/flutter_stripe/assets/349338/1c176c09-e9ce-4526-9747-df380fe574aa)

Payload with the phone field

![image](https://github.com/flutter-stripe/flutter_stripe/assets/349338/16db624a-6ea8-4eb2-acfb-b8f141ddbf61)

remonh87:
Created an issue for stripe if I have time in the weekend I will create an mr with your code

PR #1424, opened 2023-09-20
issue #1423

h3nr1ke:
Adding missing phone field to the mapper so the data will be sent into the payload

For Android and iOS

this fix a problem when using Affirm payment method (and probably some others)


remonh87:
Thank you for the fix but this needs to be fixed in the official stripe sdk because it will be overwritten for the next sync.  @jonasbark  can you confirm?

h3nr1ke:
> Thank you for the fix but this needs to be fixed in the official stripe sdk because it will be overwritten for the next sync. @jonasbark can you confirm?

Yeah, I think you are right, because they are inside the `packages` folder, and its always replaced by a newer version right?

Maybe I need to send the PR there so... I took a look and the code still the same there...

remonh87:
closing it for now :)

PR #1491, opened 2023-11-20
Sync

remonh87:
sync with Stripe sdk 0.35


packages/stripe_checkout/lib/stripe_checkout.dart
```diff
@@ -12,13 +13,17 @@ Future<CheckoutResponse> redirectToCheckout({
...
-  // ignore: deprecated_member_use_from_same_package
-  return stripe.redirectToCheckout(
-    context: context,
-    sessionId: sessionId,
-    publishableKey: publishableKey,
-    stripeAccountId: stripeAccountId,
-    successUrl: successUrl,
-    canceledUrl: canceledUrl,
-  );
+  if (kIsWeb) {
```
remonh87, line 16: @jonasbark  can you double check this change. Reason is that the conditional import does not work anymore with flutter 3.16.0 🤷‍♂️

remonh87: my implementation is wrong will rectify it

packages/stripe_platform_interface/lib/src/models/payment_methods.dart
```diff
@@ -498,6 +499,15 @@ class PaymentMethodParams with _$PaymentMethodParams {
     required PaymentMethodData paymentMethodData,
   }) = _PaymentMethodParamsPayPal;
 
+  @JsonSerializable(explicitToJson: true)
+  @FreezedUnionValue('RevoluPay')
```
jonasbark, line 503: RevolutPay*

jonasbark (approved):

PR #1495, opened 2023-11-23
Fix Inaccuracies In Documentation

GameSquatch:
Spelling issues and incorrect fields that no longer exist in the classes being used.


remonh87 (approved):
Thanks!

Issue #1467, opened 2023-11-01
collectBankAccount returns PlatformException


chilimsv:
**Describe the bug**
Calling collectBankAccount on both iOS and Android returns `PlatformException(Invalid Params, , , null)`

**To Reproduce**
Call this function 
```
 const billingDetails = BillingDetails(
        name: 'Flutter Stipe',
        email: 'email@stripe.com',
        phone: '+48888000888',
        address: Address(
          city: 'Houston',
          country: 'US',
          line1: '1459  Circle Drive',
          line2: 'S',
          state: 'Texas',
          postalCode: '77063',
        ),
      );
      await Stripe.instance.collectBankAccount(
        isPaymentIntent: true,
        clientSecret: paymentInfo.clientSecret,
        params: const CollectBankAccountParams(
          paymentMethodType: PaymentMethodType.USBankAccount,
          billingDetails: billingDetails,
        ),
      );
```
This is returned:

```
[   +1 ms] flutter: PlatformException(Invalid Params, , , null)
[        ] flutter: 
           #0      JSONMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:171:7)
           #1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:310:18)
           <asynchronous suspension>
           #2      MethodChannel.invokeMapMethod (package:flutter/src/services/platform_channel.dart:510:43)
           <asynchronous suspension>
           #3      MethodChannelStripe.collectBankAccount (package:stripe_platform_interface/src/method_channel_stripe.dart:429:20)
           <asynchronous suspension>
           #4      Stripe.collectBankAccount (package:flutter_stripe/src/stripe.dart:583:12)
           <asynchronous suspension>
```

**Expected behavior**
I expect a modal to show up to collect bank details after calling this method with required parameters

Package version: flutter_stripe: 9.5.0+1
Flutter version: stable, 3.13.0

**Additional context**
Is this the wrong usage or a bug? Every parameter is filled

xkxeeshankhan:
@chilimsv any update on the above issue?

chilimsv:
> Hey @chilimsv any update on the above issue?

No luck. Decided to collect the bank details on our client side as a workaround

lukatommy:
Hey, any update on this?  collectBankAccount still doesn't work.

I've found 2 problems:
1. Flutter plugin send 
`'isPaymentIntent':isPaymentIntent` and ios side use wrong key
[...]

PR #1507, opened 2023-11-27
fix: collectBankAccount returns PlatformException(Invalid Params)

xkxeeshankhan:
closes: #1467 

Refactor `collectBankAccount` to resolve the issue.

Changed `isPaymentIntent` to `intentType` and `BillingDetails` to `PaymentMethodData`.

The methodChannel invokes `collectBankAccount` which shows the Stripe sheet, but due to above issues the problems was occuring which is reported in #1467 .

The generated code is created using `flutter pub run build_runner build --delete-conflicting-outputs`, note that the new generated files now has a different way of creating classes.

### DIFF
--- a/packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
+++ b/packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
@@ -482,7 +482,7 @@ class MethodChannelStripe extends StripePlatform {
   }) async {
     final result = await _methodChannel
         .invokeMapMethod<String, dynamic>('collectBankAccount', {
-      'isPaymentIntent': isPaymentIntent,
+      'intentType': isPaymentIntent,
       'params': params.toJson(),
       'clientSecret': clientSecret,
     });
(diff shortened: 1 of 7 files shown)


packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
```diff
@@ -475,14 +475,13 @@ class MethodChannelStripe extends StripePlatform {
...
-  Future<PaymentIntent> collectBankAccount({
-    required bool isPaymentIntent,
+  Future<PaymentIntent> collectBankAccountPayment({
     required String clientSecret,
     required CollectBankAccountParams params,
   }) async {
     final result = await _methodChannel
         .invokeMapMethod<String, dynamic>('collectBankAccount', {
-      'isPaymentIntent': isPaymentIntent,
+      'intentType': true,
```
jonasbark, line 484: I think isPaymentIntent should remain and is a bit clearer than "intentType"

remonh87:
closed due to inactivity

PR #1510, opened 2023-11-28
fix checkout analyzer issue

remonh87:


remonh87:
@jonasbark  can you check this pr? After that I can publish a new version

packages/stripe_checkout/lib/src/platforms/stripe_checkout_mobile.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:flutter/widgets.dart';
+import 'package:stripe_checkout/src/platforms/checkout.dart';
+
+Future<CheckoutResponse?> redirectToCheckout({
```
remonh87, line 4: apparently the old method does not work anymore 🤷‍♂️  not sure if this is correct though

jonasbark (approved):

PR #1514, opened 2023-11-29
Add gradle agp8 requirement to docs

remonh87:
let me know if this is sufficient and I will publish this change


jonasbark:
I *think* this should do it - thanks!

jonasbark (approved):

PR #1547, opened 2023-12-13
chore: improvising us bank account code

rahulrepala:
The code prompts for both the descriptor code and microdeposit amount. When Stripe anticipates the descriptor code, it is essential to provide the descriptor code; otherwise, if it expects the microdeposit amount, you should provide the microdeposit amount. It is not possible to provide the microdeposit amount when the code is anticipating the descriptor code.


remonh87 (approved):
thanks!

PR #1620, opened 2024-01-29
clean up payment element, use flutter best practices, enable scrolling

siherrmann:
I updated the stateful widget to use best practices (private state and `State<PaymentElement> createState()...`). I reordered the widget parameters to be a little bit more in an order like standard flutter widgets. I updated the height so it can be set and the div inside is scrollable or it is not set and the widget gets updated by the div height inside. The constraints were not used so I deleted them.


jonasbark:
Looks good, thank you!

jonasbark (approved):

PR #1661, opened 2024-03-02
revert create checkout session on server example

jslattery26:
create checkout was not working in the example- i believe this was an issue with a sync commit about 5mos ago, i think the create-checkout-session in the api was a duplicate method of the customer method below. just reverting the changes to a previous commit because this is what works


cedvdb:
@remonh87 this is still an issue

remonh87:
checkout is not supported anymore by this library and not part of the server.  I would recommend checking: https://docs.stripe.com/payments/checkout for the integration

Issue #707, opened 2022-05-04
Not Getting Card's Fingerprint in Flutter... plugin :- flutter_stripe: ^2.5.0


jaimik11:
/// Card data associated with the payment method
class Card with _$Card {
  @JsonSerializable(explicitToJson: true)
  const factory Card({
    /// The brand associated to the card e.g. (visa, amex).
    String? brand,

    /// Two letter iso code.
    String? country,

    /// four digit number representing the year of expiry of the card.
    int? expYear,

    /// two digit number representing the month of expire of the card.
    int? expMonth,

    /// card funding type e.g. (credit, debit).
    String? funding,

    /// last four digits of the card.
    String? last4,
  }) = _Card;

  factory Card.fromJson(Map<String, dynamic> json) => _$CardFromJson(json);
}

Not Getting Card Fingerprint, How can I validate if card already exists or attached to customer? Because as of now, the customer can add many Cards with the same details. I've read in some SDKs / Library that we can check it using the payment_method's fingerprint but I'm not able to see it when I access the Card properties. How can i check card is unique or not ?

jonasbark:
This will be available soon https://github.com/stripe/stripe-react-native/blob/master/CHANGELOG.md

Flucadetena:
Was this ever add it? How can we access it in either android or IOS?

Thanks for the time and help.

Flucadetena:
Hi @jaimik11 not sure you still need this but I created a pull request to fix this issue. Hopefully it will get merged soon ;)
In the meantime you can use my fork like this. Hope it helps ;)

In your pubspec.yaml
```
  flutter_stripe:
    git:
      url: https://github.com/Flucadetena/flutter_stripe.git
      path: packages/stripe

dependency_overrides:
  stripe_platform_interface:
    git:
      url: https://github.com/Flucadetena/flutter_stripe.git
      path: packages/stripe_platform_interface
```

Issue #688, opened 2022-04-21
validate if payment method already exists using fingerprint


fernan542:
**Describe the bug**
How can I validate if card already exists or attached to customer? Because as of now, the customer can add many Cards with the same details. I've read in some SDKs / Library that we can check it using the `payment_method`'s fingerprint but I'm not able to see it when I access the `Card` properties.

Card properties:
```
/// Card data associated with the payment method
class Card with _$Card {
  @JsonSerializable(explicitToJson: true)
  const factory Card({
    /// The brand associated to the card e.g. (visa, amex).
    String? brand,

    /// Two letter iso code.
    String? country,

    /// four digit number representing the year of expiry of the card.
    int? expYear,

    /// two digit number representing the month of expire of the card.
    int? expMonth,

    /// card funding type e.g. (credit, debit).
    String? funding,

    /// last four digits of the card.
    String? last4,
  }) = _Card;

  factory Card.fromJson(Map<String, dynamic> json) => _$CardFromJson(json);
}
```

 - Package version: 2.4.0
 - Flutter version 2.10.3

Issue #1721, opened 2024-04-16
Type cast fails on Customer Sheet cancel


ovchars2:
**Describe the bug**
When a Customer Sheet is closed, a type cast error is thrown instead of StripeException

**To Reproduce**
1. Open CustomerSheetScreen
2. Init Customer Sheet
3. Present Customer Sheet
4. Close Customer Sheet by tapping above the sheet or using close button in the top right corner

**Expected behavior**
StripeException with FailureCode.Canceled is thrown

**Smartphone / tablet**
 - Device: Pixel 3
 - OS: Android 12
 - Package version: `main` branch
 - Flutter version: 3.19.3

**Additional context**
Logs:
```
W/ActivityResultRegistry(27246): Dropping pending result for request CustomerSheet: ActivityResult{resultCode=RESULT_OK, data=Intent { (has extras) }}
D/StripeReactNative(27246): com.stripe.android.customersheet.InternalCustomerSheetResult$Canceled cannot be cast to com.stripe.android.view.AddPaymentMethodActivityStarter$Result
[log] type 'String' is not a subtype of type 'Map<String, dynamic>' in type cast
      type 'String' is not a subtype of type 'Map<String, dynamic>' in type cast
      #0      _$$StripeErrorGenericImplFromJson (package:stripe_platform_interface/src/models/errors.g.dart:13:40)
      #1      new _$StripeErrorGenericImpl.fromJson (package:stripe_platform_interface/src/models/errors.freezed.dart:123:7)
      #2      _$StripeErrorFromJson (package:stripe_platform_interface/src/models/errors.freezed.dart:18:10)
      #3      new StripeError.fromJson (package:stripe_platform_interface/src/models/errors.dart:26:7)
      #4      _$$CustomerSheetResultImplFromJson (package:stripe_platform_interface/src/models/customer_sheet.g.dart:131:13)
      #5      new _$CustomerSheetResultImpl.fromJson (package:stripe_platform_interface/src/models/customer_sheet.freezed.dart:1024:7)
      #6      _$CustomerSheetResultFromJson (package:stripe_platform_interface/src/models/customer_sheet.freezed.dart:855:31)
      #7      new CustomerSheetResult.fromJson (package:stripe_platform_interface/src/models/customer_sheet.dart:100:7)
      #8      MethodChannelStripe._parseCustomerSheetResult (package:stripe_platform_interface/src/method_channel_stripe.dart:358:36)
[...]

PR #1764, opened 2024-05-09
Added "fingerprint" to the Card Model to sync with React-Native

Flucadetena:
Added the fingerprint to the card model and tested it is working correctly.
Solves Issues:
- #707 
- #688

Resolves discussion:
- #691

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,8 @@ Packages with dependency updates only:
 
 #### `stripe_platform_interface` - `v10.2.0`
 
+ - **FIX**: Not Getting Card's Fingerprint in Flutter (#707).
+ - **FIX**: validate with fingerprint (#688).
  - **FIX**: add bankidentifiercode to ideal paymentdata (#1522).
  - **FIX**: #1459 allow image to be null (#1466).
  - **FIX**: #1409 send usesDeprecatedTokenFlow to stripe platform (#1410).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,47 @@
 All notable changes to this project will be documented in this file.
 See [Conventional Commits](https://conventionalcommits.org) for commit guidelines.
 
+## 2024-05-09
+
+### Changes
+
+---
+
+Packages with breaking changes:
+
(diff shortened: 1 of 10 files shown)


Flucadetena:
After lots of testing I saw the "Type cast error String not Map<String, dynamic>" referred in issue: #1721. Was due to the CustomerSheet returning a PaymentMethod as well as an error. (To be precise, to the way the error was added to the response and as consequence handled)

To reproduce this issue you had to call the "presentCustomerSheet()" select a card and click the "confirm" button. Then call it again and close it without selecting any new option. This would cause the response to have the previously selected option and would add the "Canceled" error as part of the response.

Both on Android and IOS this situation was handled the same way. I read how other components manage this situation to see what was the preferred way of doing this. But realized that no other has this situation, as all prompt the user with a form (payment or otherwise) and if the user closes the form it will simple return a "Canceled" error, but no other info previously saved.

This is the same when the "presentCustomerSheet()" is called for the first time, with no previous selection saved, and close the sheet either by clicking outside or in the "X" icon. This case would simple return and "StripeException" with the "Canceled" code. 
The same as with any other Stripe component.

But in the "saved" selection case, it does not want to through an "StripeException" as the user may have discarded the sheet due to his "preferred" paymentMethod already being selected. So it returns a normal responses and adds the error "canceled" as an extra.
The model for this was an StripeError, but due to the process it resulted in "Type error" as it tried to access "json["code"]" when the "json" value was already the "Canceled" error.

In the normal errors, stripe creates a StripeException, that at the end create a "LocalizedErrorMessage" with the "json['code']" value. So that is what I decided to do, as thought it was the more consistent thing with the rest of the platform.

This way it will resolve successfully and return the saved "preferred" paymentMethod, and if you access the error element in the response it will have the same structure and a normal error in an StripeException.

Not sure if this explanation was necessary, sorry for the loooong text, ahhahah. But wanted to explain my decision.

Hope it helps. :)

Flucadetena:
There is an issue I can not solve, but not sure if it is the expected behavior.

When you present the CustomerSheet, add a new Card and confirm, that card won't bring the "fingerprint". The card had to be added prior to this for it to bring the fingerprint back.

PR #1775, opened 2024-05-15
Add missing example server dependencies and allow cross-origin requests when the example app runs on web.

ignatz:


jonasbark:
thank you!

PR #1794, opened 2024-06-10
feat: add payment element default values parameter

kleytonstn:
Add parameter `PaymentElementDefaultValues` in the `PaymentElement`


jonasbark:
thanks!

remonh87:
I will take this change along with the new stripe interop will create a new pr for this

Issue #1834, opened 2024-07-05
Strange error in deferred PaymentSheet in iOS


pgiacomo69:
In iOS the Credit Card Payment (deferred PaymentSheet) is unsuccesful, with a message:
"An error occurred in PaymentSheet. Your PaymentIntent setupFutureUsage (none) does not match the PaymentSheet.IntentConfiguration setupFutureUsage (Optional (StripePaymentSheet.PaymentSheet.IntentConfi guration.SetupFutureUsage.onSession))."

Apple Pay works well. Same App in Android Emulator works perfectly.
  
1. Launch the "Stripe Example" App
2. Select "Single step (differed payment)" from "payment Sheet" demos 
3. Enter the fake card data.
4. Tap the confirm button.
5. Observe the failure.

Smartphone / tablet**
 - Device: IPhone 15 Pro Emulator
 - OS: iOS 17
 - Package version: >=10.1
 - Flutter version 3.22

**Additional context**

`await Stripe.instance.initPaymentSheet(
        paymentSheetParameters: SetupPaymentSheetParameters(
          merchantDisplayName: 'Flutter Stripe Store Demo',
          allowsDelayedPaymentMethods: true,
          intentConfiguration: IntentConfiguration(

              mode: IntentMode(
                setupFutureUsage: null,
                currencyCode: 'eur',
                amount: amount,
              ),

              confirmHandler: (method, saveFuture) {
                _createIntentAndConfirmToUser(method);
              }),

          // Extra params
          primaryButtonLabel: 'Pay now',
          applePay: const PaymentSheetApplePay(
            merchantCountryCode: 'DE',
          ),
          googlePay: const PaymentSheetGooglePay(
            merchantCountryCode: 'DE',
            testEnv: true,
          ),

          style: ThemeMode.light,
          appearance: const PaymentSheetAppearance(
            colors: PaymentSheetAppearanceColors(
                /*  background: Colors.white,
              componentBackground: Colors.white,
              componentText: Colors.black,
              primary: Colors.blue,
              componentBorder: Colors.black,
              placeholderText: Colors.black,
              primaryText: Colors.black,
              secondaryText: Colors.green, */

                ),
            shapes: PaymentSheetShape(
              borderWidth: 0.5,
              borderRadius: 8,
              shadow: PaymentSheetShadowParams(color: Colors.yellow),
            ),
            primaryButton: PaymentSheetPrimaryButtonAppearance(
[...]

PR #1835, opened 2024-07-08
Fixes Issue #1834

pgiacomo69:
If the value for setupFutureUsage from json map is not null, it was incorrectly serialized as "onSession" when invoking iOS platform code. after further investigations, I found that this happens because a null value in dart, comes in platform sida as a "Optional<null>". 
I only have a doubt about this PR: It is ok as is, or it would be better to add "includeIfNull:" false to "@JsonSerializable" directives in stripe_platform_interface library?
In doubt I proposed this solution, that doesn't change anyting in platforms other than iOS.


remonh87:
@pgiacomo69  I think this pull request can be closed correct since it consist of a stripe sdk change that if we want this needs to be incorporated in the stripe_ios repo

PR #1839, opened 2024-07-09
Fix/deferred payment methods

### LINKED ISSUE #1837: Payment Methods inconsistencies when using PaymentSheet in deferred mode
When using PaymentSheet in 'Deferred Mode' the allowed payment methods can't come from intent, because the intent will be generated only after the user has choosen a payment method. Payment methods that will be shown should are those selected in Stripe Settings. From code we should be able to further filtering allowed payment methods by giving an array when istantiating `IntentConfiguration`.

With current code we have two problems:

1. In iOS, payment methods like iDeal and PayPayl won't be shown
2. In Android all configured payment methods are shown, regardless of wich values we put in `paymentMethodTypes` array in `intentConfiguration`

**To Reproduce**
1. Enable  PayPal, an other payment types in your Stripe account
2. Add `paymentMethodTypes: ['card','paypal'],` in `IntentConfiguration` at 'initPaymentSheet' method of payment_sheet_deffered_screen.dart file
3. Run the example App
4. Launch 'Single Step (deferred payment)' demo from 'Payment Sheet' group
7. In iOS: Observe that only 'Card' Payment is shown
8. In Android: Observe that all configured payment methods are shown, even if not present in `paymentMethodTypes` array.

Solution:

1. iOS: A return Url should be configured when configuring the payment sheet
2. Android: JSONArray is missing from possible tipes in ReadableMap.java from stripe_android package

A PR will be submitted to address this issue.

pgiacomo69:
Fixes #1837

### DIFF
--- a/packages/stripe_ios/ios/Classes/Stripe Sdk/StripeSdk+PaymentSheet.swift
+++ b/packages/stripe_ios/ios/Classes/Stripe Sdk/StripeSdk+PaymentSheet.swift
@@ -224,11 +224,9 @@ extension StripeSdk {
             mode = PaymentSheet.IntentConfiguration.Mode.payment(
                 amount: amount,
                 currency: modeParams["currencyCode"] as? String ?? "",
-                setupFutureUsage: { return switch modeParams["setupFutureUsage"] as? String {
-                                                      case "OffSession":  .offSession
-                                                      case "OnSession":  .onSession
-                                                      default:  nil
-                                                  } }(),
+                setupFutureUsage: modeParams["setupFutureUsage"] != nil
+                    ? (modeParams["setupFutureUsage"] as? String == "OffSession" ? .offSession : .onSession)
+                    : nil,
             captureMethod: captureMethod
             )
         } else {
--- a/packages/stripe_ios/ios/Classes/Stripe Sdk/StripeSdk+PaymentSheet.swift
+++ b/packages/stripe_ios/ios/Classes/Stripe Sdk/StripeSdk+PaymentSheet.swift
@@ -224,9 +224,11 @@ extension StripeSdk {
             mode = PaymentSheet.IntentConfiguration.Mode.payment(
                 amount: amount,
                 currency: modeParams["currencyCode"] as? String ?? "",
-                setupFutureUsage: modeParams["setupFutureUsage"] != nil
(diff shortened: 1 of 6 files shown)


remonh87:
thank you for your contribution however we do not change the underlying stripe sdk. If you want this behavior to change you should create an pr for https://github.com/stripe/stripe-react-native .

jonasbark:
We will however merge the Android part

pgiacomo69:
Hi @remonh87 , @jonasbark do you think that the new commit can solve the issue? 
Thank you 👍

packages/stripe_ios/ios/Classes/Stripe Sdk/StripeSdk+PaymentSheet.swift
```diff
@@ -224,9 +224,11 @@ extension StripeSdk {
             mode = PaymentSheet.IntentConfiguration.Mode.payment(
                 amount: amount,
                 currency: modeParams["currencyCode"] as? String ?? "",
-                setupFutureUsage: modeParams["setupFutureUsage"] != nil
-                    ? (modeParams["setupFutureUsage"] as? String == "OffSession" ? .offSession : .onSession)
-                    : nil,
+                setupFutureUsage: { return switch modeParams["setupFutureUsage"] as? String {
```
remonh87, line 227: can you remove this change and let's go for a fix in dart or we need to make a change to stripe ios

remonh87 (changes requested):

packages/stripe_ios/ios/Classes/Stripe Sdk/StripeSdk+PaymentSheet.swift
pgiacomo69: I thought to have removed it in last commit, while fixing it in dart, let me check when I return home

pgiacomo69: But, a moment, this seems the rollback, compare it against  main branch

pgiacomo69: Ok, check it now again, please

remonh87 (approved):

PR #1852, opened 2024-07-12
Add deferred payment methods for web

EduardDumitrescu:
Hello,

I've been using the library for a project and added some functionality/ fixed some bugs and thought I should try to contribute.

For features:
- added methods that would let a developer do a deferred payment on web
- exposed multiple customization options for the web elements widget
- on android we can now set the locale of a payment sheet through parameter

For bug fixes:
- on android filtering the payment methods were not working due to wrong mapping
- on ios payments like 'ideal' were not working due to wrong mapping of the field 'setupFutureUsage'


packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/PaymentSheetFragment.kt
```diff
@@ -145,7 +145,9 @@ class PaymentSheetFragment(
         putBoolean("shouldSavePaymentMethod", shouldSavePaymentMethod)
       }
 
-      stripeSdkModule.sendEvent(context, "onConfirmHandlerCallback", params)
+      context.currentActivity?.runOnUiThread {
```
remonh87, line 148: I am not convinced about this change since it will block the UI. Also this change needs to be ported to https://github.com/stripe/stripe-react-native since we copy the sdk 1 by 1

packages/stripe_js/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 3.4.1
+**Features**
+- add method for deferred payment
```
remonh87, line 3: we need to port this to the new jsinterop plugin. I am about to merge this soon.

packages/stripe_platform_interface/lib/src/models/payment_sheet.dart
```diff
@@ -106,6 +106,13 @@ class SetupPaymentSheetParameters with _$SetupPaymentSheetParameters {
     /// The list of preferred networks that should be used to process payments made with a co-branded card.
     /// This value will only be used if your user hasn't selected a network themselves.
     @JsonKey(toJson: _cardBrandListToJson) List<CardBrand>? preferredNetworks,
+
+    /// set the payment sheet language
+    String? languageCode,
```
remonh87, line 111: We need to provide the interface provided by the stripe sdk in react native and as you can see it does not have language code yet: https://github.com/stripe/stripe-react-native/blob/master/src/types/PaymentSheet.ts

packages/stripe/lib/src/stripe.dart
```diff
@@ -284,6 +284,16 @@ class Stripe {
     }
   }
 
+  Future<PaymentMethod> createPaymentMethodWithElements() =>
+      _platform.createPaymentMethodWithElements();
+
+  Future<void> elementsSubmit() => _platform.elementsSubmit();
+
+  Future<PaymentIntent> handleNextActionForNonCard(
```
remonh87, line 292: I am not convinced about the naming of this method it is very confusing for users. Why cant we use handle next action?

remonh87 (changes requested):
Thank you for your contribution but I see a couple of problems:

- we never change the stripe react native sdk folder
- let's port deferred payments for the web to the new interop (I am going to merge the branch into main today)
- I do not understand handleNextActionForNonCard this is not a normal stripe method
-  The payment sheet fields are not part of the react native sdk

Let's also split the pull request in two parts:
- deferred payment for web 
- and  the other fixes

Let me know if you need help

packages/stripe/lib/src/stripe.dart
EduardDumitrescu: In web_stripe.dart we already have a 'handleNextAction' method but it is calling js function 'handleCardAction'

```
  @override
  Future<PaymentIntent> handleNextAction(String paymentIntentClientSecret,
      {String? returnURL}) async {
    final stripe_js.PaymentIntentResponse response =
        await _stripe.handleCardAction(paymentIntentClientSecret);

    return response.paymentIntent!.parse();
  }
```

The method handleNextActionForNonCard is calling the js `handleNextAction` method. Now it would be better to rename the initial `handleNextAction` method to `handleCardAction` but that would be a breaking change on existing codebases and I didn't really know the policy of this package regarding that hence why I came up with the new name.

packages/stripe_platform_interface/lib/src/models/payment_sheet.dart
EduardDumitrescu: I understand, this was more of a feature that I added since a language force is possible on the web but not on the mobile platforms.

PR #1856, opened 2024-07-17
Feat/new interop

remonh87:
New Stripe web interop 🚀


packages/stripe_web/lib/src/widgets/card_field.dart
```diff
@@ -58,7 +58,7 @@ class WebStripeCardState extends State<WebCardField> with CardFieldContext {
     // ignore: undefined_prefixed_name
     ui.platformViewRegistry.registerViewFactory(
```
jamesblasco, line 59: This method is now defined here so we could remove the analyzer ignore

```dart
import 'dart:ui_web' as web;
web.platformViewRegistry.registerViewFactor
```

jamesblasco (approved):

PR #1866, opened 2024-07-23
sync with React Native 0.38.3

jonasbark:


remonh87 (approved):
looks good to me I do not see any thing we need to do on dart side. Will publish things later

PR #1917, opened 2024-09-08
Fix a couple issues that arose in platform pay on web after the package:web upgrade

cornwe19:
* Not calling preventDefault on PlatformPayButton means users will end up with duplicate platform pay requests
* Calling 'complete' on pay request wasn't working


cornwe19:
@remonh87 couple touchups to platform pay on web when you have time to review

jonasbark (approved):

jonasbark:
Thanks!

PR #1923, opened 2024-09-11
document android plugin issue in readme

remonh87:


jonasbark:
Can you please downgrade the example app as well?

PR #1924, opened 2024-09-11
Fix attempt to present view not in hierarchy

dluksza:
In a situation when `rootViewController` is not set by the application,
the main view controller is not found, resulting in error:

```
Attempt to present <STP_Internal_BottomSheetViewController> on <UIViewController> (from <UIViewController>) whose view is not in the window hierarchy.
````

which then prevents the Stripe bottom sheet from showing. Blocking the
payment flow.

To prevent this, a fallback is added to traverse the windows hierarchy
and find the first one that has `isKeyWindow` set to `true`.

Co-authored-by: Bruno Pastre <bruno.pastre@onefootball.com>


jonasbark:
Thanks for your contribution! This change would be overwritten from the next sync with https://github.com/stripe/stripe-react-native
Would you kindly add the same change into that repository?

dluksza:
this is now replaced by PR to `stripe-react-native` project https://github.com/stripe/stripe-react-native/pull/1736

PR #1933, opened 2024-09-19
Fix android payment sheet shapes not rendering correct when using double type

volkanakbayir:
'double' value type was not affecting the android side of rendering and creating inconsistent UI among Android's and iOS's UI rendering when tried to change payment sheet appearance. Using integer values is fixing this issue for android to actually abide this. iOS can also use the integer values as well. 

That is of course a breaking change and needs to be discussed carefully.  There maybe a better solution for this to handle it properly elsewhere like the platform code themselves? I see the code is copied from stripe-react-native I didn't want to touch that part.

Also added missing 'borderRadius' support for "PaymentSheetPrimaryButtonShape" class as this is already supported by both of the sdk's as well as in the stripe-react-native implementation.

### DIFF
--- a/packages/stripe_platform_interface/CHANGELOG.md
+++ b/packages/stripe_platform_interface/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 11.1.1
+- BREAKING: Changed payment sheet appearance related variables to use `int` instead of `double` to support android rendering. 
+- Support for PaymentSheetPrimaryButtonShape's added borderRadius added Android.
+
 ## 11.1.0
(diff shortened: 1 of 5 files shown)


jonasbark:
Thanks for the PR!

The double -> int change is not necessary. Would you mind updating your PR and change the ReadableMap.java class from:

```java
throw new Exception("We've got a double here");
```
to:
```java
return (int) getDouble(key);
```

jonasbark:
Done in https://github.com/flutter-stripe/flutter_stripe/pull/1942

PR #1971, opened 2024-10-19
Update build.gradle para adapitação da biblioteca

edu745181:
A maioria dos projetos estão exigindo essa verção do java!


remonh87:
What is the problem you want to solve? Java is backwards compatible so the plugins using an older version of java should still compile. I do not see any use for this change because we are following Google's recommended guidelines here: https://developer.android.com/build/jdks

Issue #1925, opened 2024-09-12
[Web] CardField #card-element selector applies to no DOM elements


BakerSmithA:
**Describe the bug**
Looks to be a duplicate of a previously fixed issue https://github.com/flutter-stripe/flutter_stripe/issues/1680 that happens when a `CardField` widget is wrapped in an `AnimatedOpacity` widget that starts with zero opacity. After animating from zero to one opacity, the CardField widget still does not display or allow input, and the following error is printed to the terminal:

```
IntegrationError: The selector you specified (#card-element) applies to no DOM elements that are currently on the page.
Make sure the element exists on the page before calling mount().
```

(I assume the Flutter engine is doing some optimisations and not initalising the `CardField`).

**To Reproduce**
See this repo which provides an example:

https://github.com/BakerSmithA/flutter_stripe_opacity_example

**Expected behavior**
The CardField widget allows card detail input and builds without error.

**Smartphone / tablet**
 - Device: Macbook Pro
 - OS: Sonoma
 - Package version: flutter_stripe 6.0.0, flutter_stripe_web 11.0.0
 - Flutter version 3.24.2

**Additional context**
This issue occurs after upgrading from flutter_stripe version [5.2.0](https://pub.dev/packages/flutter_stripe_web/versions/5.2.0) to [6.0.0](https://pub.dev/packages/flutter_stripe_web/versions/6.0.0), and flutter_stripe_web from [10.2.0](https://pub.dev/packages/flutter_stripe/versions/10.2.0) to [11.0.0](https://pub.dev/packages/flutter_stripe/versions/11.0.0).

remonh87:
@jamesblasco  I think we can change this widget using the same mutation observer trick that we use for paymentelement am I correct?

remonh87:
Closed by #1978

aiiro-s:
Testing to get comments

PR #1978, opened 2024-10-28
Initialize platform web pay button when added to DOM

cornwe19:
* Mimic payment element initialization


cornwe19:
@jamesblasco @jonasbark @remonh87 FYI on this one. We started using a new RUM framework and noticed a bunch of the following errors popping up when trying to use the platform pay button on mobile safari

```
IntegrationError: The selector you specified (#platform-pay-button) applies to no DOM elements
that are currently on the page. Make sure the element exists on the page before calling mount().
```

This fix addresses that issue. Similar problem to what's called out in https://github.com/flutter-stripe/flutter_stripe/issues/1925

remonh87 (approved):
thank you I wanted to do this already for a long long time

PR #1989, opened 2024-11-07
Setup confirm stripe web

cedvdb:
This effectively adds setup intent support for any kind of payment method.

This adds [ConfirmSetup](https://docs.stripe.com/js/setup_intents/confirm_setup) to stripe web as confirmSetupIntent is deprecated. I used [ConfirmPayment] as a template to follow the conventions of this package.

An example was added (and the old one removed)


cedvdb:
@remonh87 This is now ready for review

cedvdb:
integration tests failing because of 

```
StripeCore/STPAPIClient.swift:150: Fatal error: You must use a valid publishable key. For more info, see https://stripe.com/docs/keys
```

example/server/src/index.ts
```diff
@@ -369,14 +369,17 @@ app.post('/create-setup-intent', async (req, res) => {
 
   //@ts-ignore
   const setupIntent = await stripe.setupIntents.create({
-    ...{ customer: customer.id, payment_method_types },
```
remonh87, line 372: I would not change the index.ts ourselves because we copy it from stripe. I will do an update here. Can you revert this?

packages/stripe_web/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## 6.2.1
```
remonh87, line 1: Can you undo the versioning to 6.2.1 for now. I think with the new sync we probably go for a 6.3.0

remonh87 (approved):
Can you look at the 2 minor comments but for the rest it looks good! thanks for contributing!

packages/stripe_web/CHANGELOG.md
cedvdb: So should I'll remove this entirely then or change it to 6.3.0 ?

cedvdb: I changed it to 6.3.0 and in the pubspec too.

PR #2014, opened 2024-12-15
Bank accounts

remonh87:
fix bank accounts and analyzer issuez example app


example/android/settings.gradle
```diff
@@ -18,7 +18,7 @@ pluginManagement {
 
 plugins {
     id "dev.flutter.flutter-plugin-loader" version "1.0.0"
-    id "com.android.application" version "8.7.3" apply false
+    id "com.android.application" version "8.6.1" apply false
```
jonasbark, line 21: is that change on purpose?

jonasbark: @remonh87

remonh87: yes I couldn't compile the app on my android studio version but I can revert it otherwise.

jonasbark (approved):

PR #2022, opened 2024-12-23
fix: prevent focus loop in PaymentElement widget

### LINKED ISSUE #2021: [Web] Focus Loop in `PaymentElement` Causes Input Fields to Lose Focus Repeatedly.
## **Describe the bug**  
When using the `PaymentElement` widget from the `flutter_stripe_web` package, input fields lose focus instantly after being selected. This issue is observed primarily in Chrome browsers, where the `Focus` state toggles repeatedly, making the input fields unusable.

---

## **To Reproduce**  

1. Initialize the `PaymentElement` widget with the following parameters:  
   ```dart
   PaymentElement(
     onCardChanged: (details) {},
     clientSecret: secret, // secret is always there.
     width: MediaQuery.of(context).size.width,
     enablePostalCode: false,
     appearance: const ElementAppearance(
       theme: ElementTheme.night,
     ),
   ),
   ```  
2. Run `flutter run -d chrome`
3. Attempt to focus on any input field (e.g., credit card number).  
4. Observe that the input loses focus immediately after gaining focus.

https://github.com/user-attachments/assets/8d3c2391-d184-4ff2-b198-11c7e287c114

---

## **Expected behavior**  
Input fields in the `PaymentElement` widget should retain focus when selected, allowing users to enter their details seamlessly.  

---

## **Smartphone / tablet**  

- **Device**: N/A (Web platform)  
- **Browser**: Chrome  Version 131.0.6778.140
- **OS**: macOS 14.4  
- **Package version**: flutter_stripe_web: ^6.3.0 | flutter_stripe: ^11.3.0 
- **Flutter version**: 3.24.4  

---

## **Additional context**  

- This issue does not occur on Safari, where the widget works as expected.  
- The following code logs indicate continuous focus toggling:  
upon tweaking `onFocusChange` to print `focus` as follow : 
image

  ```text
  Focus Gained  
  Focus Lost  
  ``` 
  
  and upon uncommenting the commented :
  ```
  /*  if (focus)
            element?.focus();
          else
            element?.blur(); */
```
the loop happens.
  
- **Potential Cause**: Interaction between Flutter’s `FocusNode` and the underlying `iframe` behavior of the Stripe Payment Element.

### PR DESCRIPTION
**Pull request by x544D**
### What does this PR do?

This PR fixes an issue where the `PaymentElement` widget from the `flutter_stripe_web` package repeatedly toggles focus when an input field is selected, causing the input to lose focus. This issue occurred due to a feedback loop of focus/blur events.

### Changes Made:
- Added flags (`_isManuallyFocusing`, `_isCurrentlyFocused`) to prevent focus/blur recursion.
- Ensured that focus state change only occurs if there's an actual change in focus.
- Prevented unnecessary calls to `element?.focus()` and `element?.blur()` based on the new conditions.

### Testing:
- Tested the fix locally in Chrome.
- Verified that the `PaymentElement` input fields retain focus and the focus state toggling issue is resolved.

https://github.com/user-attachments/assets/11c9b722-cfe0-4d28-aa52-177947b65b6f

Fixes #2021

### CODE REVIEW

File: packages/stripe_web/lib/src/widgets/payment_element.dart [hunk 1]
```diff
@@ -169,15 +169,35 @@ class PaymentElementState extends State<PaymentElement> {
...
       focusNode: _effectiveNode,
-      onFocusChange: (focus) {
-        /*  if (focus)
-            element?.focus();
-          else
-            element?.blur(); */
+     onFocusChange: (focus) {
+        // Prevent feedback loop from manual focus/blur actions
+        if (_isManuallyFocusing) {
+          _isManuallyFocusing = false;
```
remonh87, line 182: just double checking shouldn't we call `setState` here?

x544D: Hey @remonh87, Thanks for the feedback!
No need for a `setState` As we're not updating UI.

PR #2065, opened 2025-02-06
Fix: move getClientSecret invocation after first frame

giandifra:
### Why
Trying the app on web, `Express checkout` and `Web Payment Element` has infinite loading due to an exception from `ScaffoldMessenger.of(context);` of `getClientSecret` method.

### What
Invoke `getClientSecret` after first frame. 

```
@override
  void initState() {
    super.initState();
    SchedulerBinding.instance.addPostFrameCallback((_) {
      getClientSecret();
    });
  }
```

### Video

https://github.com/user-attachments/assets/e22fda96-fe7c-4be9-af4a-99c5640b6855


remonh87 (approved):
thanks

PR #2078, opened 2025-02-16
Fix #2072 throw an exception instead of a crash for unknown type plat…

### LINKED ISSUE #2072: Error: type '_Map<String, dynamic>' is not a subtype of type 'bool?' in type cast
**Describe the bug**
There is a good number of Firebase Crashlytics getting fired when we try ti call `.isPlatformPaySupported`. This seems to happen to Android only.

StackTrace:
```
Error: type '_Map<String, dynamic>' is not a subtype of type 'bool?' in type cast, StackTrace: #1335 MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:334) <asynchronous suspension>
#1572 MethodChannelStripe.isPlatformPaySupported
(package:stripe_platform_interface/src/method_channel_stripe.dart:428
<asynchronous suspension> #2027 
Stripe.isPlatformPaySupported (package:flutter_stripe/src/stripe.dart:135 <asynchronous suspension>
```

We are trying to use `Stripe.instance.isPlatformPaySupportedListenable.value` where value is expecting a bool, yet for some reason, we are getting `_Map<String, dynamic>`. I assume there is an exception happening and we can try catch it, but nothing is standing out as to why this error is coming from the library?

**To Reproduce**
Haven't been able to reproduce ourselves

**Expected behavior**
No crash alerts coming into crashlytics.

**Smartphone / tablet** - OS: Android 13 & 14
 - Package version: 11.1.0
 - Flutter version 3.24.3

**Additional context**

remonh87:
…form pay supported

### DIFF
--- a/packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
+++ b/packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
@@ -424,10 +424,16 @@ class MethodChannelStripe extends StripePlatform {
   }) async {
     bool? isSupported;
     if (params == null) {
-      isSupported =
-          await _methodChannel.invokeMethod('isPlatformPaySupported', {
+      final result =
+          await _methodChannel.invokeMethod<dynamic>('isPlatformPaySupported', {
         'params': {},
       });
+
+      if (result is bool) {
+        isSupported = result;
+      } else {
+        StripeException.fromJson(result);
+      }
     } else {
       isSupported = await _methodChannel
           .invokeMethod('isPlatformPaySupported', {'params': params.toJson()});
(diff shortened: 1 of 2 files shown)


packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
```diff
@@ -424,10 +424,20 @@ class MethodChannelStripe extends StripePlatform {
...
-          await _methodChannel.invokeMethod('isPlatformPaySupported', {
+      final result =
+          await _methodChannel.invokeMethod<dynamic>('isPlatformPaySupported', {
         'params': {},
       });
+
+      if (result is bool) {
+        isSupported = result;
+      } else {
+        throw StripeException(
```
jonasbark, line 435: That would then result in a different exception rather than a fix. Did you find out what the result contains if not bool?

remonh87: sorry forgot to force push 😂

PR #2101, opened 2025-03-16
Sync

remonh87:


jonasbark (approved):

jonasbark:
15d06dc fix: cardfield methods

@remonh87 this change was on purpose as Stripe changed it as well. Can you double check?

remonh87:
> [15d06dc](https://github.com/flutter-stripe/flutter_stripe/commit/15d06dcfc91f24c039d610b848765ab7a3a44ce3) fix: cardfield methods
> 
> @remonh87 this change was on purpose as Stripe changed it as well. Can you double check?

@jonasbark  yes I see that on Android but not on iOS as there it is still oncardchange: 

    func onCardChange(_ arguments: Dictionary<AnyHashable, Any>?) {
        channel.invokeMethod("onCardChange", arguments: arguments)
    }

so shall we then fix it in Swift in our factory to keep it consistent?

PR #2170, opened 2025-07-07
fixed: incorrect to register cancel callback in PaymentRequest

howmind:
When cancel Google Pay in chrome, never show the native payment sheet again.
After use the correct invoke, the issue has gone.


remonh87 (approved):
thanks

Issue #2180, opened 2025-07-17
collectBankAccountToken failing due to missing parameter.


rafaelortizzableh:
**Describe the bug**
Calling collectBankAccountToken fails due to missing params parameter when calling the Method Channel. An Invalid Parameters Exception is thrown.

**To Reproduce**
1. Get a financial connections session client secret with a secret key. 
2. Pass that client secret to the `collectBankAccountToken` method as a parameter. Passing the `params` parameter is optional, but the failures occurs whether it's present or not.
3. Observe a failure with exception `PlatformException` with the following stack trace:
```
#0      JSONMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:168:7)
#1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:370:18)
<asynchronous suspension>
#2      MethodChannel.invokeMapMethod (package:flutter/src/services/platform_channel.dart:570:43)
<asynchronous suspension>
#3      MethodChannelStripe.collectBankAccountToken (package:stripe_platform_interface/src/method_channel_stripe.dart:559:20)
<asynchronous suspension>
 ... 
```

**Expected behavior**
The method is called correctly and the modal sheet that allows user to connect a bank is show.

**Smartphone / tablet**
 - Device: iPhone 13 Pro
 - OS: iOS 18
 - Package version: 11.5.0
 - Flutter version 3.32.6 (stable)

**Additional context**
After forking and adding the missing parameter, the method gets called correctly, but the `userInterfaceStyle: UserInterfaceStyle.alwaysLight` property under params is not respected.

remonh87:
merged the pr (thanks!) we will ship it in the next release hopefully coming week

PR #2181, opened 2025-07-17
fix: add missing param to collectBankAccountToken method channel

rafaelortizzableh:
Should fix the main issue reported on #2180.


remonh87 (approved):
thanks

PR #2188, opened 2025-08-02
Sync

remonh87:
Biggest sync since years


remonh87:
@jonasbark  I am thinking before shipping lets add support for freezed as well since it is a breaking release

alexodus:
> @jonasbark I am thinking before shipping lets add support for freezed as well since it is a breaking release. I can work on it let me know your thoughts

with freezed 3.2.0 map/when are back.

remonh87:
> > @jonasbark I am thinking before shipping lets add support for freezed as well since it is a breaking release. I can work on it let me know your thoughts
> 
> with freezed 3.2.0 map/when are back.

that is good news! will save a lot of hassle.

PR #2202, opened 2025-08-22
Changing Stripe packages

fareed069417:


remonh87:
Thank you for contributing but unfortunately I have to reject this changeset. All files in the reactnativesdk folder are a 1 on 1 copy of the react native stripe sdk. This allows us for a risk free and relatively quick sync. By accepting this changeset each sync with the stripe official sdk will be more painful.

PR #2258, opened 2025-11-11
Prefer `initSurfaceAndroidView` instead of `initExpensiveAndroidView`

lreardon:
initSurfaceAndroidView is a newer method that tries to avoid expensive rendering methods unless necessary. It increases performance significantly on older devices and prevents intermittent crashes.

## Summary by CodeRabbit

* **New Features**
  * Added new Android rendering option for payment form components, providing an alternative surface-based platform view implementation.

* **Refactor**
  * Updated Android platform view rendering strategy across payment widgets (CardField, CardFormField, AddressSheet, AUBECSDebitForm, GooglePayButton) to use surface-based rendering for improved performance and consistency.


remonh87:
@jonasbark  what do you think shall we take this change along with the new sync? I think it can dramatically improve the performance

PR #2264, opened 2025-11-18
[FIX] CollectBankAccountToken

maksym-eshyft:
Move the created workaround logic within `collectBankAccountToken` to execute only after confirming no error occurred.

This prevents an access error that occurs when a user cancels the bank account collection flow, as the result object is incomplete and attempting to read the created property throws an exception.


maksym-eshyft:
@remonh87 @jonasbark

PR #2267, opened 2025-11-24
Feat: web language

flt-phy:
**Scope**: As a user I want to be able to change language of Stripe web elements on the go, without refreshing the page completely (PaymentElement for Card, Paypal, ApplePay, GooglePay)

**Previous behaviour**: browser's default language is used.

**Implemented functionality**: allow setting `locale` and forwarding it into stripe web (there was already `locale` variable, but it was not used)

Current implementation works only on Web, though param is visible for Android/iOS implementations (but with appropriate disclaimer comment).

image


remonh87:
Thank you

PR #2312, opened 2026-01-26
fix: 2306 remove descendantsAreFocusable are focusable as it dismisse…

remonh87:
…s the keyboard when we do not want it


jonasbark:
same as https://github.com/flutter-stripe/flutter_stripe/pull/2316 but the other one has more context

PR #2319, opened 2026-01-29
feat: add separate flutter_stripe_identity package

### LINKED ISSUE #1819: Support Stripe identity verification
Support for the Stripe identity verification embedded element would be nice. React native supports it.

realmeylisdev:
## Summary

This PR adds **Stripe Identity Verification** support as a **separate federated plugin**, following the maintainer's guidance in #2318 and the pattern used by [stripe-identity-react-native](https://github.com/stripe/stripe-identity-react-native).

Closes #1819

## Why Separate Package?

Per @jonasbark's feedback on #2318:
> "We currently synchronize our code base with the one from Stripe React Native. Most of your Android / iOS changes would be overwritten then."

This approach:
- **Won't be overwritten** - Native code lives in separate packages
- **Mirrors Stripe's pattern** - Same as `stripe-identity-react-native`
- **Independent versioning** - Can be released/updated separately
- **Smaller bundle** - Apps not using Identity don't include it

## Package Structure

```
packages/
├── stripe_identity/                          # Main facade (flutter_stripe_identity)
├── stripe_identity_platform_interface/       # Abstract interface + models
├── stripe_identity_ios/                      # iOS native (StripeIdentity SPM 25.0.1)
├── stripe_identity_android/                  # Android native (com.stripe:identity:22.2.+)
└── stripe_identity_web/                      # Web stub (throws WebUnsupportedError)
```

## API Usage

```dart
import 'package:flutter_stripe_identity/flutter_stripe_identity.dart';

final result = await StripeIdentity.instance.presentIdentityVerificationSheet(
  verificationSessionId: 'vs_xxx',
  ephemeralKeySecret: 'ek_xxx',
);

switch (result) {
  case IdentityVerificationCompleted():
    print('Verification completed');
  case IdentityVerificationCanceled():
    print('User canceled');
  case IdentityVerificationFailed(:final error):
    print('Error: ${error.message}');
}
```

## Key Differences from #2318

| Aspect | PR #2318 | This PR |
|--------|----------|---------|
| Approach | Modify existing packages | Separate packages |
| Method channel | `flutter.stripe/payments` | `flutter.stripe_identity/identity` |
| Native code | In synced files | Standalone plugins |
| React Native compat | Uses RN shims | Clean native code |

## Test Plan

- [ ] iOS: Identity sheet presents correctly
- [ ] iOS: Verification flow completes
- [ ] Android: Identity sheet presents correctly  
- [ ] Android: Verification flow completes
- [ ] Web: Throws `WebUnsupportedError`
- [ ] Example app works on iOS/Android

### Testing Requirements
- Backend endpoint that creates VerificationSession
- Identity enabled in Stripe Dashboard
- Real device recommended (camera-based verification)
- Camera permissions configured

## Summary by CodeRabbit

* **New Features**
  * Added Stripe Identity verification support to Flutter with a unified API for presenting the verification sheet on iOS and Android.
  * Added structured handling for completed, canceled, and failed outcomes, including error details.
  * Included a full Flutter example app demonstrating the verification flow and branding.
* **Documentation**
  * Added READMEs for the main plugin, platform implementations, and the shared interface.
* **Tests**
  * Added a basic widget test for the example app.


packages/stripe_identity_platform_interface/pubspec.yaml
```diff
@@ -0,0 +1,24 @@
+name: stripe_identity_platform_interface
+description: Platform interface for Stripe Identity Verification
+version: 1.0.0
```
remonh87: maybe for the first version we should consider 0.1.0? What do you think @jonasbark ?

remonh87 (commented):
Looks good thank you this is a great addition some tasks that still needs to be done before we can ship it:

- [ ] Add readme to all the packages 
- [ ] I do not think we should keep a stripe_identity_web package as this would suggest that we support web which we do not. Let's remove it and as a service we can throw in stripe identity package if someone calls it from the web.

I also would like to let @jonasbark  check it from the native bindings side

packages/stripe_identity/pubspec.yaml
```diff
@@ -0,0 +1,36 @@
...
+  flutter: ">=3.0.0"
+
+flutter:
+  plugin:
+    platforms:
+      android:
+        default_package: stripe_identity_android
+      ios:
+        default_package: stripe_identity_ios
+      web:
```
remonh87: let's not add web support if we cannot promise something working

remonh87 (changes requested):

rlee1990:
Any updates on this?

remonh87 (approved):
lookgs good to me thanks for applying the fixes. I will cook some CI runners and will make sure we will launch it

remonh87:
@jonasbark  I think we can integrate it but would be good if you can double check it.

rlee1990:
Hi, just checking in to see if this will still be implemented

packages/stripe_identity_platform_interface/pubspec.yaml
realmeylisdev: Done — all four packages (`flutter_stripe_identity`, `stripe_identity_platform_interface`, `stripe_identity_ios`, `stripe_identity_android`) are set to `0.1.0` for the initial release, per your suggestion.

packages/stripe_identity/pubspec.yaml
realmeylisdev: Agreed — I removed the separate `stripe_identity_web` package entirely and replaced it with a `kIsWeb` guard in the main package that throws `UnsupportedError`, so there is no half-working web surface. Web is also documented as unsupported in the README.

packages/stripe_identity/example/lib/screens/identity_verification_screen.dart
```diff
@@ -0,0 +1,127 @@
...
+      );
+
+      final data = jsonDecode(response.body);
+
+      if (data['error'] != null) {
+        throw Exception(data['error']);
+      }
+
+      final sessionId = data['id'] as String;
+      final ephemeralKeySecret = data['ephemeral_key_secret'] as String;
```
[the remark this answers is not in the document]
realmeylisdev: Fixed in 6675d1c1 — the example now checks the HTTP status code before decoding and parses the response body as a typed `Map<String, dynamic>`.

packages/stripe_identity/example/lib/widgets/loading_button.dart
```diff
@@ -0,0 +1,61 @@
...
+    } catch (e, s) {
+      log(e.toString(), error: e, stackTrace: s);
+      scaffoldMessenger.showSnackBar(SnackBar(content: Text('Error $e')));
+      rethrow;
+    } finally {
+      setState(() {
+        _isLoading = false;
+      });
+    }
+  }
```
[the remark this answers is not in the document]
realmeylisdev: Fixed in 6675d1c1 — the `finally` block now guards `setState` with a `mounted` check.

packages/stripe_identity/example/lib/widgets/loading_button.dart
```diff
@@ -0,0 +1,61 @@
...
+      _isLoading = true;
+    });
+    final scaffoldMessenger = ScaffoldMessenger.of(context);
+
+    try {
+      await widget.onPressed!();
+    } catch (e, s) {
+      log(e.toString(), error: e, stackTrace: s);
+      scaffoldMessenger.showSnackBar(SnackBar(content: Text('Error $e')));
+      rethrow;
```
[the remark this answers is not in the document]
realmeylisdev: Fixed in 6675d1c1 — dropped the `rethrow`. The error is already logged and surfaced via the SnackBar, and this handler is not awaited by the caller, so nothing needs to propagate.

PR #2324, opened 2026-02-01
fix: send empty dict instead of null for confirmPayment params

realmeylisdev:
## Summary
- Fixes Apple Pay `confirmPayment` by sending empty dict `{}` instead of `null` for params
- Resolves issue where Swift guard rejects null params, breaking Apple Pay flow

## Problem
Swift requires non-null `NSDictionary` for params in `StripePlugin.swift`, but Dart sends `null` when Apple Pay is used (payment method is created by `STPApplePayContext`, not from Dart).

## Solution
Send empty dict instead of null, matching the existing pattern used for `options`.

## Test plan
- [ ] Test Apple Pay confirmPayment flow
- [ ] Test regular card payment confirmPayment flow

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved robustness of payment confirmation by ensuring proper handling of null payment parameters, guaranteeing consistent serialization of payment data.

### DIFF
--- a/.github/workflows/all_plugins.yaml
+++ b/.github/workflows/all_plugins.yaml
@@ -138,10 +138,18 @@ jobs:
     name: integration test (Android)
     needs: [analyze, test]
     timeout-minutes: 60
-    runs-on: macos-15
+    runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v4
     - uses: subosito/flutter-action@v2
+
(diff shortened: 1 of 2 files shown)


remonh87:
@meylis1998 I am not sure if this problem fixes the issue. The issue is that the `PaymentMethodParams ` does not have an empty constructor. See also the discussion on https://github.com/flutter-stripe/flutter_stripe/pull/2287

PR #2337, opened 2026-02-16
Card field delay

russellsjensen:
Use MutationObserver to wait for the DOM element to be inserted, matching the pattern used by PaymentElement. This is necessary because postFrameCallback timing is unreliable in deployed environments so sometimes the card field would not render.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Stripe card field initialization with enhanced timing handling for better cross-browser compatibility, particularly in Safari and post-frame rendering scenarios.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -63,3 +63,5 @@ lcov.info
 /test-results/
 /playwright-report/
 /playwright/.cache/
+
+**/build/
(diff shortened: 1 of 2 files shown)


remonh87:
Closed due to no reaction feel free to reopen it with updated comments

PR #2372, opened 2026-03-18
fix(web): unique view type per PaymentElement instance to fix infinite spinner on remount (#2368)

### LINKED ISSUE #2368: [Web] PaymentElement shows infinite loading spinner on remount (regression in 7.4.0)
**Describe the bug**

Since `7.4.0`, the `PaymentElement` widget displays an infinite `CircularProgressIndicator` when
the screen containing it is navigated away from and then navigated back to (i.e. the widget is
disposed and remounted). The form never becomes interactive. Rolling back to `7.3.0` resolves the
issue immediately.

The root cause is in the new `_mountWhenConnected` polling loop introduced for WASM support.
`platformViewRegistry.registerViewFactory` silently ignores re-registrations for an already-known
view type ID (`'stripe_payment_element'`). On the second mount, the new `_divElement` instance is
never actually injected into the DOM, so `_divElement.isConnected` is always `false`.
`_mountWhenConnected` then loops infinitely via `addPostFrameCallback`, `_isReady` is never set to
`true`, and the `CircularProgressIndicator` overlay is never removed.

**To Reproduce**

1. Initialise Stripe and render a `PaymentElement` inside a screen:

```dart
class PaymentScreen extends StatefulWidget {
  const PaymentScreen({super.key, required this.onBack});
  final VoidCallback onBack;
  @override
  State<PaymentScreen> createState() => _PaymentScreenState();
}

class _PaymentScreenState extends State<PaymentScreen> {
  bool _initialized = false;

  @override
  void initState() {
    super.initState();
    WebStripe.instance
        .initialise(
          publishableKey: '<your_publishable_key>',
          merchantIdentifier: 'merchant.example',
        )
        .then((_) {
          if (mounted) setState(() => _initialized = true);
        });
  }

  @override
  Widget build(BuildContext context) {
    return Scaffold(
      appBar: AppBar(leading: BackButton(onPressed: widget.onBack)),
      body: _initialized
          ? PaymentElement(
              clientSecret: '<your_payment_intent_client_secret>',
              customerSessionClientSecret: '<your_ephemeral_key>',
              onCardChanged: (_) {},
            )
          : const SizedBox.shrink(),
    );
  }
}
```

2. Navigate to this screen — `PaymentElement` loads correctly.
3. Press the back button — the widget is disposed.
4. Navigate to the screen again.
5. Observe that a `CircularProgressIndicator` is displayed indefinitely and the payment form
   never appears.

**Expected behavior**

`PaymentElement` should load and become fully interactive on every mount, exactly as it does in
`7.3.0`.

**Smartphone / tablet**

- Device: Desktop / browser (Flutter Web only)
- OS: macOS 15 / Chrome 145
- Package version: `flutter_stripe_web 7.4.0` (regression vs `7.3.0`)
- Flutter version: 3.29.x

**Additional context**

I think this is quite critical as a user going back and forth between screens containing a `PaymentElement` will not be able to use the payment form.

Viotty:
## Summary

- Fixes #2368 

When navigating away from a screen containing a `PaymentElement` and back to it, the widget
displayed an infinite `CircularProgressIndicator` and the payment form never became interactive.

### Route cause

`platformViewRegistry.registerViewFactory` silently ignores re-registrations for an already-known view type ID.
- The hardcoded ID `'stripe_payment_element'` was registered on the first `initState`. 
- On the second mount, a new `_divElement` is created but its factory registration is ignored. Flutter's renderer keeps using the stale factory from the first mount. 
- The new div is never injected into the DOM, so `_divElement.isConnected` stays `false` forever.
- `_mountWhenConnected` then loops infinitely via `addPostFrameCallback`, `_isReady` is never set to `true`, and the `CircularProgressIndicator` overlay is never removed.

### Fix

Generate a unique view type ID per state instance using a static counter (`stripe_payment_element_0`, `stripe_payment_element_1`, …). Each mount registers a fresh factory, so `HtmlElementView` always resolves to the correct div for that instance.

### Test plan

- [ ] Navigate to a screen containing `PaymentElement` => form loads correctly
- [ ] Navigate away (widget disposed) then navigate back => form loads correctly on remount

## Summary by CodeRabbit

* **Chores**
  * Improved internal handling for multiple payment element instances.

### DIFF
--- a/packages/stripe_web/lib/src/widgets/payment_element.dart
+++ b/packages/stripe_web/lib/src/widgets/payment_element.dart
@@ -72,6 +72,9 @@ class PaymentElement extends StatefulWidget {
 }
 
 class PaymentElementState extends State<PaymentElement> {
+  static int _nextId = 0;
+  final String _viewType = 'stripe_payment_element_${_nextId++}';
+
   web.HTMLDivElement _divElement = web.HTMLDivElement();
   double height = 300.0;
   bool _isReady = false;
@@ -108,7 +111,7 @@ class PaymentElementState extends State<PaymentElement> {
     _cachedElementOptions = _elementOptionsOnce();
 
     ui.platformViewRegistry.registerViewFactory(
-      'stripe_payment_element',
+      _viewType,
       (int viewId) => _divElement,
     );
 
@@ -176,7 +179,7 @@ class PaymentElementState extends State<PaymentElement> {
         ),
         child: Stack(
           children: [
-            const HtmlElementView(viewType: 'stripe_payment_element'),
+            HtmlElementView(viewType: _viewType),
             if (!_isReady)
               Container(
                 color: Theme.of(context).scaffoldBackgroundColor,


cedvdb:
shouldn't there be tests so it does not happen again ?

Issue #2378, opened 2026-03-30
Sync v0.61 left overs


remonh87:
We want to move fast but we lack a couple of things that needs to be added natively and in the dart layer. For more info see: https://github.com/stripe/stripe-react-native/compare/v0.58.0...v0.61.0

- add createRadarSession to the plugin
- Create the paymenetmethodmessaging component viewfactories + dart implementation
- Implement the Stripe connectviews

jonasbark:
- fix termsDisplay logic

remonh87:
closed in v

PR #2385, opened 2026-04-06
fix: correct termsDisplay type to Map<String, TermsDisplay>

realmeylisdev:
## Summary
- Fix `termsDisplay` field type from `TermsDisplay?` (single enum) to `Map<String, TermsDisplay>?`
- Both native platforms expect a map like `{"card": "automatic"}` but the Dart model was sending a bare string, causing it to be silently ignored

Closes part of #2378

## Test plan
- [ ] Verify serialized JSON contains map structure
- [ ] Run `initPaymentSheet` with termsDisplay map on both platforms

## Summary by CodeRabbit

* **Improvements**
  * Payment sheet setup now supports configurable terms-display settings per payment method, giving finer control over how terms are shown.
* **Bug Fixes**
  * Unknown or unsupported terms-display values now safely default to automatic presentation to ensure consistent behavior.


remonh87 (approved):
thank you That was indeed wrong

PR #2395, opened 2026-04-14
fix!: return sealed CollectBankAccountResult from collectBankAccount / verifyMicrodeposits

BiAksoy:
## Summary

- **Fixes** `collectBankAccount` and `verifyPaymentIntentWithMicrodeposits` when called with `isPaymentIntent: false`. The native SDKs return a setup-intent-shaped payload that cannot be parsed by `PaymentIntent.fromJson` (missing `amount`, `currency`, `captureMethod`, `confirmationMethod`).
- **Introduces** a sealed `CollectBankAccountResult` union with `.paymentIntent(PaymentIntent)` / `.setupIntent(SetupIntent)` variants.
- Both `MethodChannelStripe` methods now delegate to `ResultParser<T>` + the generated `fromJson` for both branches, removing all hand-written field extraction and status mapping.

---

## Breaking change

`Stripe.instance.collectBankAccount` and `Stripe.instance.verifyPaymentIntentWithMicrodeposits` now return `Future<CollectBankAccountResult>` instead of `Future<PaymentIntent>`. 

Callers must now destructure the union:

```dart
final result = await Stripe.instance.collectBankAccount(...);

switch (result) {
  case CollectBankAccountPaymentIntentResult(:final paymentIntent):
    // handle payment intent flow
    break;
  case CollectBankAccountSetupIntentResult(:final setupIntent):
    // handle setup intent flow
    break;
}
```

**Note:** The setup-intent branch was broken before this change, so no working code depends on the previous shape.

---

## Affected packages

- `stripe_platform_interface`
- `stripe`
- `stripe_web`

## Summary by CodeRabbit

* **Breaking Changes**
  * The return type of `collectBankAccount()` and `verifyPaymentIntentWithMicrodeposits()` has changed from a generic intent to a sealed union result type that explicitly indicates whether a payment or setup intent was obtained. Code calling these methods requires updates to properly handle the distinct result types and their associated data.

### DIFF
--- a/packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
+++ b/packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
@@ -15,6 +15,7 @@ import 'package:stripe_platform_interface/src/result_parser.dart';
 
 import 'models/app_info.dart';
 import 'models/card_details.dart';
+import 'models/collect_bank_account_result.dart';
 import 'models/errors.dart';
 import 'models/payment_intents.dart';
 import 'models/payment_methods.dart';
@@ -569,7 +570,7 @@ class MethodChannelStripe extends StripePlatform {
   }
 
   @override
-  Future<PaymentIntent> collectBankAccount({
+  Future<CollectBankAccountResult> collectBankAccount({
     required bool isPaymentIntent,
     required String clientSecret,
     required CollectBankAccountParams params,
@@ -583,13 +584,11 @@ class MethodChannelStripe extends StripePlatform {
 
     _financialConnectionsEventHandler = params.onEvent;
 
-    return ResultParser<PaymentIntent>(
-      parseJson: (json) => PaymentIntent.fromJson(json),
-    ).parse(result: result!, successResultKey: 'paymentIntent');
+    return CollectBankAccountResult.fromJson(result!);
   }
 
   @override
-  Future<PaymentIntent> verifyPaymentIntentWithMicrodeposits({
+  Future<CollectBankAccountResult> verifyPaymentIntentWithMicrodeposits({
     required bool isPaymentIntent,
     required String clientSecret,
     required VerifyMicroDepositsParams params,
@@ -601,9 +600,7 @@ class MethodChannelStripe extends StripePlatform {
           'clientSecret': clientSecret,
         });
 
-    return ResultParser<PaymentIntent>(
-      parseJson: (json) => PaymentIntent.fromJson(json),
-    ).parse(result: result!, successResultKey: 'paymentIntent');
+    return CollectBankAccountResult.fromJson(result!);
   }
 
   @override
--- a/packages/stripe_platform_interface/lib/src/models/collect_bank_account_result.dart
+++ b/packages/stripe_platform_interface/lib/src/models/collect_bank_account_result.dart
@@ -1,0 +1,38 @@
+import 'package:freezed_annotation/freezed_annotation.dart';
+
+import 'errors.dart';
+import 'payment_intents.dart';
+import 'setup_intent.dart';
+
+part 'collect_bank_account_result.freezed.dart';
+
+/// Result of `collectBankAccount` / `verifyPaymentIntentWithMicrodeposits`.
+///
+/// Native SDKs return either a payment intent or a setup intent depending on
+/// the client secret that was supplied, so the result is modelled as a sealed
+/// union to keep both flows strongly typed.
+@freezed
+sealed class CollectBankAccountResult with _$CollectBankAccountResult {
+  const factory CollectBankAccountResult.paymentIntent(
+    PaymentIntent paymentIntent,
+  ) = CollectBankAccountPaymentIntentResult;
+
+  const factory CollectBankAccountResult.setupIntent(SetupIntent setupIntent) =
+      CollectBankAccountSetupIntentResult;
+
+  /// Parses a native method-channel response into the matching variant, or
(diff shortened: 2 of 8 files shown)


packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
```diff
@@ -601,9 +600,73 @@ class MethodChannelStripe extends StripePlatform {
...
+  }) {
+    if (isPaymentIntent) {
+      return ResultParser<PaymentIntent>(
+        parseJson: (json) => PaymentIntent.fromJson(json),
+      ).parse(result: result, successResultKey: 'paymentIntent');
+    }
+
+    final setupIntentJson = result['setupIntent'] as Map<String, dynamic>?;
+    if (setupIntentJson != null) {
+      return PaymentIntent(
```
remonh87: I do not like this this we should always rely on automatic parsing of the map through freezed  + jsonserializable like we do above.

packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
```diff
@@ -601,9 +600,73 @@ class MethodChannelStripe extends StripePlatform {
...
+    }
+
+    // Fallback: try paymentIntent key in case the native SDK format changes.
     return ResultParser<PaymentIntent>(
       parseJson: (json) => PaymentIntent.fromJson(json),
-    ).parse(result: result!, successResultKey: 'paymentIntent');
+    ).parse(result: result, successResultKey: 'paymentIntent');
+  }
+
+  static PaymentIntentsStatus _parseStatus(String value) {
```
remonh87: this should not be needed when we parse it using freezed + json serializable

remonh87 (changes requested):
I do not think this is correct if the key is different we should handle this but not create our own parsing ourselves

BiAksoy:
Thanks, that makes sense.

I added the manual parsing because the native response is setup-intent-shaped and doesn’t fully match PaymentIntent.fromJson, but I agree this isn’t the right layer for it.

I’ll refactor this to rely on the existing freezed + json_serializable parsing and align with the current pattern. 

For the setup intent case, I may introduce a sealed result type to keep both flows consistent. Let me know if you’d prefer a different approach.

I can also update the feat/web PR (#2396) in the same way if that makes sense.

BiAksoy:
Pushed the refactor.

packages/stripe_platform_interface/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+## 13.0.0
```
remonh87: Can you revert this change? We will update the changelogs when we actually release.

packages/stripe_platform_interface/lib/src/models/collect_bank_account_result.dart
```diff
@@ -0,0 +1,21 @@
...
+
+part 'collect_bank_account_result.freezed.dart';
+
+/// Result of `collectBankAccount` / `verifyPaymentIntentWithMicrodeposits`.
+///
+/// Native SDKs return either a payment intent or a setup intent depending on
+/// the client secret that was supplied, so the result is modelled as a sealed
+/// union to keep both flows strongly typed.
+@freezed
+sealed class CollectBankAccountResult with _$CollectBankAccountResult {
```
remonh87, line 15: I think if you add here JSONSerializable you do not need the parser later.

packages/stripe_web/CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## 8.0.0
```
remonh87: Can you revert this change? We will update the changelogs when we actually release.

packages/stripe/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: flutter_stripe
 description: Flutter library for Stripe. Supports PaymentSheets, Apple & Google Pay, SCA, PSD2 and much more.
-version: 12.6.0
+version: 13.0.0
```
remonh87: also remove versioning

packages/stripe_web/pubspec.yaml
```diff
@@ -1,6 +1,6 @@
 name: flutter_stripe_web
 description: Stripe sdk bindings for the Flutter web platform. This package contains the implementation of the platform interface for web.
-version: 7.6.0
+version: 8.0.0
```
remonh87: also remove versioning

remonh87 (changes requested):
Almost there! If you can revert the changelog and versioning we can ship this.  You can think about the jsonserialzable instead of the resultparser but this is not required.

BiAksoy:
Pushed.

- Reverted the changelog and version bump changes.
- On the parser point: moved the dispatch out of MethodChannelStripe into a CollectBankAccountResult.fromJson factory on the union itself, so the call site is just CollectBankAccountResult.fromJson(result!) and the inner parsing still flows through the json_serializable-generated PaymentIntent.fromJson / SetupIntent.fromJson.

I didn't put @JsonSerializable directly on the union because our native payload uses the key name itself as the discriminator ({"paymentIntent": ...} vs {"setupIntent": ...}) rather than a separate type field, so freezed's auto generated union dispatch doesn't fit cleanly. The manual fromJson is ~10 lines and still delegates all the field work to json_serializable. Happy to restructure if you'd prefer the annotation.

remonh87 (approved):
Nice thank you

PR #2400, opened 2026-04-22
Fix: #2382 confirm token handler on payment sheet

remonh87:
Fixes #2382 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed confirmation token parsing in payment callbacks so the correct token field is used.
  * Improved payment sheet initialization to properly detect and enable available confirmation handlers, ensuring handlers are invoked as expected.

### DIFF
--- a/packages/stripe_ios/ios/stripe_ios/Sources/stripe_ios/StripePlugin.swift
+++ b/packages/stripe_ios/ios/stripe_ios/Sources/stripe_ios/StripePlugin.swift
@@ -447,16 +447,9 @@ extension  StripePlugin {
 
     func initPaymentSheet(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
         guard let arguments = call.arguments as? FlutterMap,
-              var params = arguments["params"] as? NSDictionary else {
+              let params = arguments["params"] as? NSDictionary else {
             result(FlutterError.invalidParams)
             return
-        }
-        if (params.object(forKey: "intentConfiguration") != nil && params.object(forKey: "intentConfiguration") is NSDictionary) {
-            let mutable = (params["intentConfiguration"] as! NSDictionary).mutableCopy() as! NSMutableDictionary
-            mutable["confirmHandler"] = true;
-            let adjusted = params.mutableCopy() as! NSMutableDictionary
-            adjusted["intentConfiguration"] = mutable
-            params = adjusted
         }
         initPaymentSheet(params: params, resolver: resolver(for: result), rejecter: rejecter(for: result))
     }
--- a/packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
+++ b/packages/stripe_platform_interface/lib/src/method_channel_stripe.dart
@@ -80,7 +80,7 @@ class MethodChannelStripe extends StripePlatform {
           _confirmTokenHandler != null) {
         final method = ResultParser<ConfirmationTokenResult>(
           parseJson: (json) => ConfirmationTokenResult.fromJson(json),
-        ).parse(result: call.arguments!, successResultKey: 'paymentMethod');
+        ).parse(result: call.arguments!, successResultKey: 'confirmationToken');
         _confirmTokenHandler!(method);
       } else if (call.method == 'onCustomPaymentMethodConfirmHandlerCallback' &&
           _confirmCustomPaymentMethodCallback != null) {
@@ -243,8 +243,18 @@ class MethodChannelStripe extends StripePlatform {
   Future<PaymentSheetPaymentOption?> initPaymentSheet(
     SetupPaymentSheetParameters params,
   ) async {
+    final paramsJson = params.toJson();
+    final intentConfig = paramsJson['intentConfiguration'];
+    if (intentConfig is Map<String, dynamic>) {
+      if (params.intentConfiguration?.confirmHandler != null) {
+        intentConfig['confirmHandler'] = true;
+      }
+      if (params.intentConfiguration?.confirmTokenHandler != null) {
+        intentConfig['confirmationTokenConfirmHandler'] = true;
+      }
+    }
     final result = await _methodChannel.invokeMethod('initPaymentSheet', {
-      'params': params.toJson(),
+      'params': paramsJson,
     });
     if (params.intentConfiguration?.confirmHandler != null) {
       _confirmHandler = params.intentConfiguration?.confirmHandler;


packages/stripe_ios/ios/stripe_ios/Sources/stripe_ios/StripePlugin.swift
```diff
@@ -453,7 +453,9 @@ extension  StripePlugin {
         }
         if (params.object(forKey: "intentConfiguration") != nil && params.object(forKey: "intentConfiguration") is NSDictionary) {
             let mutable = (params["intentConfiguration"] as! NSDictionary).mutableCopy() as! NSMutableDictionary
-            mutable["confirmHandler"] = true;
+            if mutable["confirmationTokenConfirmHandler"] == nil {
```
remonh87: This fix is to prevent that we do not run into the error you must provide either `confirmationTokenConfirmHandler` or `confirmationTokenConfirm` but not both

packages/stripe_ios/ios/stripe_ios/Sources/stripe_ios/StripePlugin.swift
```diff
@@ -447,17 +447,10 @@ extension  StripePlugin {
 
     func initPaymentSheet(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
         guard let arguments = call.arguments as? FlutterMap,
-              var params = arguments["params"] as? NSDictionary else {
+              let params = arguments["params"] as? NSDictionary else {
             result(FlutterError.invalidParams)
             return
         }
-        if (params.object(forKey: "intentConfiguration") != nil && params.object(forKey: "intentConfiguration") is NSDictionary) {
```
remonh87, line 454: this is not needed anymore: Now that Dart explicitly sends exactly the right flag  (confirmHandler or confirmationTokenConfirmHandler), the iOS layer should just pass the params through  unchanged.

scott-pestshare:
> **Actionable comments posted: 1**
> 
> Caution
> 
> Some comments are outside the diff and can’t be posted inline due to platform limitations.
> 
> ⚠️ Outside diff range comments (1)
> > packages/stripe_ios/ios/stripe_ios/Sources/stripe_ios/StripePlugin.swift (1)> `202-209`: _⚠️ Potential issue_ | _🔴 Critical_
> > > **Match the confirmation-token completion payload key used by Dart.**
> > > `MethodChannelStripe.confirmationTokenCreationCallback` sends `params`, but iOS only reads `result`, so after `_confirmTokenHandler` is invoked the completion callback is rejected as `invalidParams` on iOS.
> > > 
> > > 🐛 Proposed fix
> > > ```diff
> > >          case "confirmationTokenCreationCallback":
> > > -            guard let arguments = call.arguments as? FlutterMap,
> > > -                  let res = arguments["result"] as? NSDictionary else {
> > > +            guard let arguments = call.arguments as? FlutterMap else {
> > > +                result(FlutterError.invalidParams)
> > > +                return
> > > +            }
> > > +            guard let res = (arguments["result"] as? NSDictionary)
> > > +                    ?? (arguments["params"] as? NSDictionary) else {
> > >                  result(FlutterError.invalidParams)
> > >                  return
> > >              }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > 
> > > 
> > > 🤖 Prompt for AI Agents
> > > ```
> > > Verify each finding against the current code and only fix it if needed.
> > > 
> > > In `@packages/stripe_ios/ios/stripe_ios/Sources/stripe_ios/StripePlugin.swift`
> > > around lines 202 - 209, The iOS handler for the
> > > "confirmationTokenCreationCallback" case is reading the wrong payload key
> > > ("result") causing invalidParams; update the guard in StripePlugin.swift (inside
> > > the switch case for "confirmationTokenCreationCallback") to read let res =
> > > arguments["params"] as? NSDictionary (keeping the existing call.arguments as?
> > > FlutterMap guard) and pass that res into
> > > confirmationTokenCreationCallback(result: res) so the iOS key matches the Dart
> > > MethodChannelStripe.confirmationTokenCreationCallback payload.
> > > ```
> 
> 🤖 Prompt for all review comments with AI agents
> ```
> Verify each finding against the current code and only fix it if needed.
> 
> Inline comments:
> In `@packages/stripe_ios/ios/stripe_ios/Sources/stripe_ios/StripePlugin.swift`:
> - Around line 456-458: The code incorrectly synthesizes
> mutable["confirmHandler"] = true when mutable["confirmationTokenConfirmHandler"]
> is nil, which can mark a confirm handler as present even when Dart didn't
> provide one; update the logic in StripePlugin.swift where mutable is constructed
> (look for keys "confirmationTokenConfirmHandler" and "confirmHandler" and the
> downstream onConfirmHandlerCallback usage) so that you only set "confirmHandler"
> when an explicit Dart flag for it exists (or when
> confirmationTokenConfirmHandler is truthy), otherwise leave it unset/null to
> allow native “missing handler” validation to catch absent callbacks.
> 
> ---
> 
> Outside diff comments:
> In `@packages/stripe_ios/ios/stripe_ios/Sources/stripe_ios/StripePlugin.swift`:
> - Around line 202-209: The iOS handler for the
> "confirmationTokenCreationCallback" case is reading the wrong payload key
> ("result") causing invalidParams; update the guard in StripePlugin.swift (inside
> the switch case for "confirmationTokenCreationCallback") to read let res =
> arguments["params"] as? NSDictionary (keeping the existing call.arguments as?
> FlutterMap guard) and pass that res into
> confirmationTokenCreationCallback(result: res) so the iOS key matches the Dart
> MethodChannelStripe.confirmationTokenCreationCallback payload.
> ```
> 
> 🪄 Autofix (Beta)
> Fix all unresolved CodeRabbit comments on this PR:
> 
> * [ ]   Push a commit to this branch (recommended)
> * [ ]   Create a new PR with the fixes
> 
> ℹ️ Review info
> ⚙️ Run configuration
> **Configuration used**: defaults
> 
> **Review profile**: CHILL
> 
> **Plan**: Pro
> 
> **Run ID**: `89f71c4d-1382-4854-a74d-2866df89e85d`
> 
> 📥 Commits
> Reviewing files that changed from the base of the PR and between [21cf701](https://github.com/flutter-stripe/flutter_stripe/commit/21cf7016af83beb99097b064a545df269fcc07fa) and [c744634](https://github.com/flutter-stripe/flutter_stripe/commit/c7446347660cba4ba3a548db13025d5d2d06362e).
> 
> 📒 Files selected for processing (2)

I'm finding that this feedback is valid and needs to be applied, as InvalidParams is otherwise always thrown.

PR #2401, opened 2026-04-22
fix(android): crash when app backgrounded during Google Pay readiness check

simpson-peter:
## Summary
- Fix crash in `GooglePayLauncherManager` when the app is backgrounded while Google Pay availability is being checked
- `GooglePayLauncher` checks availability asynchronously; if the activity moves to `STOPPED` before the callback fires, calling `ActivityResultLauncher.launch()` throws `IllegalStateException` (enforced by AndroidX since activity 1.2.0)
- Guard `onGooglePayReady` with a lifecycle state check before calling `presentForPaymentIntent` / `presentForSetupIntent`; return `GooglePayErrorType.Failed` if the activity is not `STARTED`

### DIFF
--- a/packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/GooglePayLauncherManager.kt
+++ b/packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/GooglePayLauncherManager.kt
@@ -1,6 +1,7 @@
 package com.reactnativestripesdk
 
 import android.annotation.SuppressLint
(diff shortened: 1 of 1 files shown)


remonh87:
Thank you for the contribution but I would recommend to create the pr in: https://github.com/stripe/stripe-react-native . We do a frequent (almost) 1:1 sync with the native sdk and thus your changes will be gone next time. We also created better contributing guidelines to help you next time. Apologies.

Issue #2398, opened 2026-04-22
Link option still visible in PaymentSheet even after using LinkDisplayParams (Flutter)


hardik-cloudus:
### 🐛 Issue Description

I am using `flutter_stripe` PaymentSheet in my Flutter app.

Even after setting:

```dart
linkDisplayParams: LinkDisplayParams(linkDisplay: LinkDisplay.never),
```

The **Link (email)** option is still visible in the PaymentSheet UI.

---

### ✅ What I Tried

* Set `allowsDelayedPaymentMethods: false`
* Added `LinkDisplayParams(linkDisplay: LinkDisplay.never)`
* Backend configured with:

```json
{ "payment_method_types": ["card"], 
```

---

### ❌ Expected Behavior

Link option should be hidden from PaymentSheet.

---

### ❗ Actual Behavior

Link option is still visible and cannot be removed.

---

### 📱 Environment

* Flutter version: 3.41.6
* flutter_stripe version: 12.6.0
* Platform: Android / iOS

---

### 📸 Screenshots

<img width="358" height="454" alt="Image" src="https://github.com/user-attachments/assets/1fbd3070-c1b2-4db2-8f98-655c445c5f0e" />

It seems `LinkDisplayParams` is either not supported or ignored in PaymentSheet. Please clarify if this is expected behavior or a bug.

realmeylisdev:
Root cause: Dart serializes `{ "linkDisplayParams": { "linkDisplay": "never" } }` but native iOS (`StripeSdkImpl+PaymentSheet.swift:41`, `+Embedded.swift:321`) and Android (`PaymentElementConfig.kt:70`, called from `PaymentSheetManager.kt:309`) read `params["link"]["display"]`. Native falls through to `.automatic`, so Link keeps showing.

Already fixed on `main` by #2407 (Dart-side `@JsonKey(name: 'link')` / `@JsonKey(name: 'display')`). It's not in a pub.dev release yet — and per the maintainer's note on the closed #2406, the next release will be a breaking-change version (so 12.7.0, not a 12.6.1 patch).

**Workaround until 12.7.0 ships** — add to your app's `pubspec.yaml`:

```yaml
dependency_overrides:
  stripe_platform_interface:
    git:
      url: https://github.com/flutter-stripe/flutter_stripe
      ref: 8b8d0ebe
      path: packages/stripe_platform_interface
```

[...]

PR #2406, opened 2026-04-28
Fix native LinkDisplayParams handling

najeira:
Dart's `LinkDisplayParams` is converted to JSON as follows:

```json
{
  "linkDisplayParams": {
    "linkDisplay": "never"
  }
}
```

However, the Swift and Kotlin implementations expected the following JSON:

```json
{
  "link": {
    "display": "never"
  }
}
```

The referenced fields did not match.

This Pull Request fixes the fields referenced by Swift and Kotlin to `linkDisplayParams` and `linkDisplay`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed link display configuration parameter reading in payment sheet and embedded payment element initialization on Android and iOS platforms, ensuring accurate link display behavior during payment setup.

### DIFF
--- a/packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/PaymentElementConfig.kt
+++ b/packages/stripe_android/android/src/main/kotlin/com/reactnativestripesdk/PaymentElementConfig.kt
@@ -65,7 +65,7 @@ internal fun buildLinkConfig(params: ReadableMap?): PaymentSheet.LinkConfigurati
     return PaymentSheet.LinkConfiguration()
   }
 
-  val display = mapStringToLinkDisplay(params.getString("display"))
+  val display = mapStringToLinkDisplay(params.getString("linkDisplay"))
 
   return PaymentSheet.LinkConfiguration(
     display = display,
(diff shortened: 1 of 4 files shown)


najeira:
Fixes https://github.com/flutter-stripe/flutter_stripe/issues/2398

remonh87 (changes requested):
We should not change this in the reactnativesdk but in the platforminterface. The contents of this folder will be synced 1:1 with stripes native sdk and each of the changes will be overwritten in the next sync

najeira:
Thank you for checking this.

Just to confirm, are you suggesting changing the key name on the Dart side?

For example, like this:

```dart
@JsonKey(name: 'link')
LinkDisplayParams? linkDisplayParams,

@JsonKey(name: 'display')
required LinkDisplay linkDisplay,
```

remonh87:
> Thank you for checking this.
> 
> Just to confirm, are you suggesting changing the key name on the Dart side?
> 
> For example, like this:
> 
> ```dart
> @JsonKey(name: 'link')
> LinkDisplayParams? linkDisplayParams,
> 
> @JsonKey(name: 'display')
> required LinkDisplay linkDisplay,
> ```

Normally I would say manipulate it in the map but we can change it like you suggested because we will introduce a breaking change anyway in the next release

najeira:
The new fix will be implemented using a different way than this Pull Request, so I will close this one.

Issue #2282, opened 2025-12-19
Apple Pay no longer works with confirmPayment()


luanalbineli:
**Describe the bug**
After we upgraded the lib's version from `11.3.0` to `12.0.2` we're no longer able to use `confirmPayment()` to complete the purchase using Apple Pay. It now throws a PlatformException with the `Invalid Params` message.

**To Reproduce**
1. Launches the Apple Pay wallsheet to create the payment intent:
```dart
final paymentMethod = await Stripe.instance
    .createPlatformPayPaymentMethod(
      params: PlatformPayPaymentMethodParams.applePay(
        applePayParams: ApplePayParams(
          cartItems: [
            ApplePayCartSummaryItem.immediate(
              label: 'Label',
              amount: '10.90',
            ),
          ],
          merchantCountryCode: 'IT',
          currencyCode: 'EUR',
        ),
      ),
    );
```
2. Create the payment intent to get the secret.
3. Confirm the payment:
``` dart
await Stripe.instance.confirmPayment(
  paymentIntentClientSecret: 'secret_from_previous_step',
)
```
3. Observe a failure with exception:
```sh
PlatformException(Invalid Params, , , null) - #0      JSONMethodCodec.decodeEnvelope (package:flutter/src/services/message_codecs.dart:168:7)
#1      MethodChannel._invokeMethod (package:flutter/src/services/platform_channel.dart:367:18)
<asynchronous suspension>
#2      MethodChannel.invokeMapMethod (package:flutter/src/services/platform_channel.dart:567:43)
<asynchronous suspension>
#3      MethodChannelStripe.confirmPayment (package:stripe_platform_interface/src/method_channel_stripe.dart:132:20)
<asynchronous suspension>
#4      Stripe.confirmPayment (package:flutter_stripe/src/stripe.dart:374:29)
```

**Expected behavior**
The payment to be successfully processed using Apple Pay.  

**Smartphone / tablet**
 - Device: iPhone 9 (it should happen with any device)
 - OS: iOS 16 (same from device)
 - Package version: 12.0.2
 - Flutter version 3.35.3

**Additional context**
The `confirmPayment()` method on Swift seems to be validating if the method channel is sending the `params` argument on the most recent versions:
```swift
func confirmPayment(_ call: FlutterMethodCall, result: @escaping FlutterResult) {
    guard let arguments = call.arguments as? FlutterMap,
[...]

PR #2415, opened 2026-05-13
fix(stripe_ios): allow confirmPayment without params (possible regression)

nerder:
## Summary

This restores the iOS `confirmPayment` behavior where `params` can be omitted when the PaymentIntent already has a payment method attached server-side.

`PaymentMethodParams?` is still nullable in the Dart platform interface, and `MethodChannelStripe.confirmPayment` still sends `params: null` when `Stripe.instance.confirmPayment(...)` is called without `data`. In `stripe_ios` 12.x, the iOS method-channel handler rejects that call before the existing native server-attached-payment-method path can run.

Related to #2282, #2287, and the older #1246.

## Reproduction

This fails on iOS with `flutter_stripe >= 12.x` when confirming a PaymentIntent whose payment method is attached by the backend.

```dart
final paymentMethod = await Stripe.instance.createPlatformPayPaymentMethod(
  params: PlatformPayPaymentMethodParams.applePay(
    applePayParams: ApplePayParams(
      merchantCountryCode: 'IT',
      currencyCode: 'EUR',
      cartItems: [
        ApplePayCartSummaryItem.immediate(
          label: 'Test purchase',
          amount: '10.90',
        ),
      ],
    ),
  ),
);

// Send paymentMethod.id to the backend.
// Backend creates a PaymentIntent with this payment method attached
// and returns the PaymentIntent client secret.
await Stripe.instance.confirmPayment(
  paymentIntentClientSecret: clientSecret,
);
```

Actual result:

```txt
PlatformException(Invalid Params, , , null)
```

Expected result:

`confirmPayment` should continue into the native Stripe confirmation flow because this flow is valid when the payment method was attached server-side.

## Regression Cause

The native iOS method-channel handler currently requires `params`:

```swift
let params = arguments["params"] as? NSDictionary
```

inside the `guard`, so a Dart call with `params: null` fails immediately with `FlutterError.invalidParams`.

Older `11.x` behavior treated `params` as optional. `StripeSdkImpl.getPaymentMethodType(params: nil)` already supports this case and leads to:

```swift
STPPaymentIntentParams(clientSecret: paymentIntentClientSecret)
```

This PR restores that optionality at the Flutter method-channel boundary and in the native `confirmPayment` signature.

## Testing

- Added regression coverage in `packages/stripe_ios/test/stripe_ios_test.dart`.
- Verified in a consuming Flutter app with saved card and Apple Pay test-mode purchases.
- Ran `flutter test test/stripe_ios_test.dart` in `packages/stripe_ios`.

## Release Notes

* **Changes**
  * Stripe iOS SDK's payment confirmation method now allows optional payment parameters, providing greater flexibility in payment processing workflows.

* **Tests**
  * Added test coverage to validate optional parameter handling and method signature behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/flutter-stripe/flutter_stripe/pull/2415)

### DIFF
--- a/packages/stripe_ios/ios/stripe_ios/Sources/stripe_ios/Stripe Sdk/StripeSdkImpl.swift
+++ b/packages/stripe_ios/ios/stripe_ios/Sources/stripe_ios/Stripe Sdk/StripeSdkImpl.swift
@@ -880,7 +880,7 @@ public class StripeSdkImpl: NSObject, UIAdaptivePresentationControllerDelegate {
     @objc(confirmPayment:data:options:resolver:rejecter:)
     public func confirmPayment(
         paymentIntentClientSecret: String,
-        params: NSDictionary,
+        params: NSDictionary?,
         options: NSDictionary,
         resolver resolve: @escaping RCTPromiseResolveBlock,
         rejecter reject: @escaping RCTPromiseRejectBlock
@@ -888,10 +888,8 @@ public class StripeSdkImpl: NSObject, UIAdaptivePresentationControllerDelegate {
         self.confirmPaymentResolver = resolve
         self.confirmPaymentClientSecret = paymentIntentClientSecret
 
-        // Handle React Native null values - when null is passed from JS, it becomes NSNull
-        let actualParams = (params == NSNull()) ? nil : params
-        let paymentMethodData = actualParams?["paymentMethodData"] as? NSDictionary
-        let (missingPaymentMethodError, paymentMethodType) = getPaymentMethodType(params: actualParams)
+        let paymentMethodData = params?["paymentMethodData"] as? NSDictionary
+        let (missingPaymentMethodError, paymentMethodType) = getPaymentMethodType(params: params)
         if missingPaymentMethodError != nil {
             resolve(missingPaymentMethodError)
             return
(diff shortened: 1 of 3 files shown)


remonh87:
This pr is in violation with our [contributing guidelines ](https://github.com/flutter-stripe/flutter_stripe/blob/HEAD/CONTRIBUTING.md) because these files should not be edited. Also it is ok to use AI but please check the changes that are being made. Closing it now

nerder:
Hey @remonh87, first of all sorry for not reading the contributing guidelines before issuing the PR.

I've done this change to fix an active bug we had in prod, and I thought would be good to PR the fix. What's the suggested approach here for changes on the iOS SDK?

I'm not extremely familiar with Swift so of course that should be considered as an early dirty patch to be discussed with experts before even remotely thinking of merging it. 

I'm willing to do the work if that's helpful or simply opening an issue with a repro and let whoever you think is best suited to address it.

Let me know.

remonh87:
normally if iOS needs to be fixed it needs to be done in the [official stripe sdk](https://github.com/stripe/stripe-react-native). Then we can take it along with the next sync with their sdk

PR #2466, opened 2026-09-06
fix(android): CardField focus and IME — stop mirroring native focus, hand back text input, release focus, tap targets, keyboard on gap taps

gazialankus:
-----
- [x] Have you followed the guidelines in our [Contributing](https://github.com/flutter-stripe/flutter_stripe/blob/HEAD/CONTRIBUTING.md) document?
- [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/flutter-stripe/flutter_stripe/pulls) for the same change?
- [x] Have you added an explanation of what your changes do and why you'd like us to include them?
- [x] Do all checks pass?
- [x] Have you developed the feature on the latest stable version of Flutter?
-----
- [x] AI was used to generate or assist with generating this PR. *Please specify below how you used AI to help you, and what steps you have taken to manually verify the changes*.
-----

## Android `CardField`: five focus / IME fixes

All five share one principle: **the native card view owns its own focus and IME; the framework must never mirror or fight it.** They are ordered so each commit builds on the previous one; please review in order.

| # | Symptom (Android) | Cause | Fix |
|---|---|---|---|
| 1 | Tapping "Name" (a Flutter field) bounced focus back into the card number | `_handlePlatformFocusChanged` re-requested focus for the CardField's node on every native focus report | Drop the mirror; tap-into-card is already handled by the `Listener` |
| 2 | After leaving and returning to the card field, typed characters were dropped (backspace still worked) | `PlatformViewLink` sends `TextInput.setPlatformViewClient` from an internal `FocusNode` that can never focus inside `CardField` (`descendantsAreFocusable: false`), so the IME stayed bound to the last Flutter field | `CardField` sends `setPlatformViewClient` itself on focus, deferred to end of frame (the Flutter field being left sends `clearClient` in the same frame) |
| 3 | A caret kept blinking in the card widget after leaving it; leaving the **CVC** carried its number pad into the next field | `blur()` → `requestBlurFromJS` is hardcoded to the card *number* EditText; a focused expiry/CVC was never cleared | Send the view's `clearFocus` (clears whichever child holds focus, dropping the IME) when framework focus leaves |
| 4 | Tapping the visible CVC/expiry put the caret in the card number | The `Listener`'s focus change sent `focus()` → `requestFocusFromJS`, hardcoded to the card number | Record that a focus change came from a pointer and skip the native focus call in that case |
| 5 | Tapping the gaps between subfields, or any tap after dismissing the keyboard, focused a subfield but raised no keyboard | (a) `Listener` only requested focus when not already focused; (b) the keyboard request ran before the native focus event landed, so `findFocus()` was null | Request the keyboard when the native view reports its focused subfield (400 ms timer as the only fallback); adds a `showKeyboard` command to `stripe_android` (`showSoftInput` on the already-focused view) — `focus` cannot be reused as it is hardcoded to the card number |

## Verification

Manually on a Samsung SM-A235F (Android 14) with Flutter 3.44.0, against a production app: each row's symptom reproduced before and confirmed fixed after; iOS re-verified unchanged (every change is Android-only). `flutter analyze` and the `stripe` package tests pass.

**Reviewer note on #5:** it fixes a timing race, and scripted `adb input tap` verification *passes* on the unfixed code while real finger taps fail. Please verify with real taps, or check the ordering in a trace, not with a scripted tap.

## AI disclosure

The diagnosis and patches were developed with an AI coding assistant (Claude Code). Every symptom above was reproduced and re-tested by hand on a physical device by the author, before and after each change; the race in #5 in particular was caught only by manual testing after a scripted check had wrongly passed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QapcPohD4yibyfT5reG9jM

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved card field focus and keyboard behavior across Android and iOS.
  * The on-screen keyboard now appears more reliably when users tap already-focused card input fields.
  * Prevented unnecessary focus changes while showing the keyboard.
  * Improved focus cleanup when leaving the card field, particularly on Android.
  * Added more reliable keyboard handling for the currently focused card input.

### DIFF
--- a/packages/stripe/lib/src/widgets/card_field.dart
+++ b/packages/stripe/lib/src/widgets/card_field.dart
@@ -632,13 +632,37 @@ class _MethodChannelCardFieldState extends State<_MethodChannelCardField>
       setState(() {});
     }
     if (!isFocused) {
-      blur();
+      _releaseNativeFocus();
 
       return;
     }
 
     focus();
     _claimPlatformViewTextInput();
+  }
+
+  /// Releases the native card widget's focus when framework focus leaves.
+  ///
+  /// On Android [blur] is not enough. It lands on `requestBlurFromJS`, which
+  /// is hardcoded to the card *number* field: it clears focus on that one
+  /// EditText and then hands focus to the widget's own container. So when the
+  /// expiry or CVC field is the focused one, its focus is never cleared, and
+  /// either way Android focus stays inside the card widget. The result is a
+  /// caret still blinking in the card field while the user types in another
+  /// Flutter field, and an IME session that keeps the subfield's keyboard
+  /// configuration — most visibly the CVC's number pad following focus into a
+  /// name field.
+  ///
+  /// `clearFocus` clears the whole platform view (`cardView.clearFocus()`
+  /// clears whichever descendant holds focus) and drops the IME. It is Android
+  /// only: the iOS platform view implements just `focus`/`blur`/`clear`, and
+  /// there focusing another view already cancels the previous focus.
+  void _releaseNativeFocus() {
+    if (defaultTargetPlatform == TargetPlatform.android) {
+      _methodChannel?.invokeMethod('clearFocus');
+      return;
+    }
+    blur();
   }
 
   /// Points the engine's text input at the card field's platform view.
--- a/packages/stripe/lib/src/widgets/card_field.dart
+++ b/packages/stripe/lib/src/widgets/card_field.dart
@@ -1,3 +1,4 @@
+import 'dart:async';
 import 'dart:developer' as dev;
 
 import 'package:flutter/foundation.dart';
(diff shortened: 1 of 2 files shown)


packages/stripe/lib/src/widgets/card_field.dart
```diff
@@ -619,12 +690,125 @@ class _MethodChannelCardFieldState extends State<_MethodChannelCardField>
...
+  /// "Name on card" and come back and the card field silently drops every
+  /// character, while backspace still works, because raw key events go
+  /// straight to the focused Android view instead of through the input
+  /// connection.
+  ///
+  /// Deferred to the end of the frame on purpose: the Flutter field losing
+  /// focus sends `TextInput.clearClient`, which resets the engine to no
+  /// target. Claiming the platform view synchronously here would be undone by
+  /// that message landing afterwards.
+  void _claimPlatformViewTextInput({bool showKeyboard = false}) {
```
remonh87: the parameter show keyboard is not used

packages/stripe/lib/src/widgets/card_field.dart
```diff
@@ -346,6 +347,39 @@ class _MethodChannelCardFieldState extends State<_MethodChannelCardField>
...
+  /// Safety net for [_keyboardRequestPending], in case the native view never
+  /// reports a focused subfield. Deliberately later than that event: firing it
+  /// eagerly (from the post-frame callback, say) consumes the request before
+  /// the native focus lands and puts back the very race this avoids.
+  Timer? _keyboardRequestTimeout;
+
+  void _scheduleKeyboardRequest() {
+    _keyboardRequestPending = true;
+    _keyboardRequestTimeout?.cancel();
+    _keyboardRequestTimeout = Timer(const Duration(milliseconds: 400), () {
```
remonh87, line 375: I do not like we create a timer here because this will leak on when the view get's disposed.

packages/stripe/lib/src/widgets/card_field.dart
```diff
@@ -453,7 +487,22 @@ class _MethodChannelCardFieldState extends State<_MethodChannelCardField>
...
             widget.focusNode.requestFocus();
+          } else {
+            // Already focused, so no framework focus change will fire and
+            // nothing else would raise the keyboard. The native widget only
+            // shows it when the touch lands on one of its EditTexts, so a tap
+            // on the gaps between the fields — or any tap after the user
+            // dismissed the keyboard — would leave a focused card field with
+            // no keyboard at all.
+            _scheduleKeyboardRequest();
+            _requestNativeKeyboard();
```
remonh87: we request it twice once here and also further in line 659.

packages/stripe/lib/src/widgets/card_field.dart
remonh87: I am really concerned about this method because it reimplements internal flutter engine logic and it feels very fragile to do this. Also I am not sure if older versions of the Flutter engine support everything here.  

Instead I would: Stop blocking descendant focus on CardField's node and instead forward focus into the platform view's own internal (framework-owned)   focus node, so Flutter's built-in _onFocusChange → setPlatformViewClient plumbing claims text input for you.

This is a bigger rework that needs to be done.

remonh87 (changes requested):
Thank you for your contribution. I think the fix is sound, but it is not the direction I would like to go in because we call a couple of internal Flutter methods here. We cannot guarantee this will work in the future and also for previous Flutter engines. 

It also masks problems that we have with the architecture of the card field. I would like to think about a way to rework the rendering here to better comply with the internal workings of the engine, so we can leave more work to the framework rather than constantly needing to create hacks for it.  I will create a ticket for this.

gazialankus:
Thanks for the thorough review. Pushed 609b7b97 for the three nits. Understood on the direction — happy to test a focus-forwarding rework against the reproduction in the PR description once the ticket exists; our device runs showed the failure only under a real-tap timing race, so I can validate on hardware.

